PR #94, opened 2020-02-12
Fix scala 2.12 build 

cchepelov:
## Summary
The scala 2.12 build was broken (tests were failing due to a sheared *jackson* library set: some bits were on 2.6.7 as specified, some bits were pulled to 2.9.5 by the scala 2.12 edition of Spark 2.4.0

## Details
This splits the Jackson version, leaving it at 2.6.7 on scala 2.11 and pulling it to 2.9.6 on scala 2.12.

Probably as a result of the Jackson version increase, a couple schema test datasets had to be split betwen the scala 2.11 and the scala 2.12 builds. This is due to the fact that Jackson's JSON serializer no longer outputs something when the `indexMapping` field of `c.e.c.s.model.Type` is None.

Tests now appear to pass correctly while running sbt on the command line, but still contain a 'random-ish' failure while running in parallel mode within IDEA — to be addressed separately.


hayssams:
+1

Issue #95, opened 2020-02-12
[BUG] - cannot safely run unit tests in parallel


cchepelov:
## Description
Attempting to run unit tests in parallel hits several race conditions, which renders a "green" result a lucky occurrence in the wrong environment

## Expected behavior
Running `sbt test` or any other equivalent (such as running tests under IDEA) succeeds all the time

## Current Behavior
Running tests under IDEA on a multicore machine may result in tests randomly failing.

## Leads pursued so far

* A first-order approach of the problem is that multiple unit test instances may attempt to call  `HdfsStorageHandler#write` method on the same file target (e.g. `/metadata/types/types.yml`), and this method contains a race condition.
   * The race condition in `HdfsStorageHandler#write` is that it first deletes the destination path (if it exists) then proceeds with creating a new file, then fills its contents. If Unit Test Instance #1 attempts to write `/metadata/types/types.yml` at roughly the same time as Unit Test Instance #2 does the same, this can result in UTI #1 believing the file is now written, attempting to read back from it, while UTI #2 begins its own "delete then overwrite process".
   * An 'obvious fix' for this is to have the `HdfsStorageHandler#write` first write into a uniquely named file, then `rename()` from the temporary path to the target path using the `Rename.Options.OVERWRITE` flag (this requires to use Hadoop's FileContext rather than FileSystem)
   * This 'obvious fix' fails as while an HDFS FileSystem implementation must guarantee that the Create, Close, Delete and Rename operations are atomic, the *rename-and-overwrite* operation is *not* atomic. Indeed, it is implemented by AbstractFileSystem as the sequence of a delete then a rename-without-OVERWRITE. This contains effectively the same race condition as the initial issue.
   * Even with implementing a "keep trying" approach to solve the delete-and-rename race, this only serves to expose another race: every unit test begins by destroying every single file in the Comet data directory (`cleanMetadata` and `cleanDatasets` methods). Which means that every Unit Test begins by destroying any files needed by already-runing unit tests, dooming them.

* A second-order approach of the problem would be to modify TestHelper so that each test suite instance receives a distinct `tempPath`, effectively insulating each test unit instance from each other.
   * This fails, as the domain descriptors in `test/resources/sample/*.yml` refer to absolute paths *(one of them appears to be rooted in @hayssams own /Users directory)* ⇒ *a possible solution to this would include performing some replacements in `TestHelper#init` while pulling the resource into the storageHandler.*
[...]

Issue #96, opened 2020-02-24
[FEATURE] - support delivering audits to JDBC


cchepelov:
## Description
Audits & Metrics can be delivered to BigQuery for now. 
It would be useful to support delivering to a JDBC database instead

Work is already in progress (@hayssams)

cchepelov:
Got _a_ something that seems to push into JDBC, but #95 is now firmly in my way to test that we got exactly what we expect to into the Audit table.

hayssams:
Done

PR #98, opened 2020-02-25
[CDP-95] Fix test dataset race

cchepelov:
## Summary
This PR solves the race condition between various test suites, by removing the shared & mutable configuration singleton and putting it on the stack instead (there is still only one instance in production code, and as many instances as there are test suites)

Related issue: #95 

PR Type: Bug fix

Status: WIP

Breaking change? *MAYBE* (eventually, Yes)

### Solution
This PR moves the instances formerly within the `Settings`object into a `Settings` instance. For now, the old `Settings.comet`, `Settings.storageHandler`, `Settings.schemaHandler` lazy variables are (at least temporarily) transformed into trampolines that keep the former syntax available (at least as long as an implicit `Settings` instance is around)

As a result of this PR:
* each test suite runs in its own directory, insulated from any other instance (but still sharing the global `SparkSession` instance)
* each test suite may now define programmatic overrides on the configuration used to load the `Settings`. This may be useful in #96 for instance, to enable testing with/without JDBC indexing.

### How has this been tested?
`sbt clean test` is green. (**EDIT**: it was green but actually failing, due to files remaining in my `/tmp` which are no longer delivered there. Thanks, Travis)

### Other changes

When under test, LockFile's polling period is now reduced to 5 milliseconds (down from 5000 milliseconds), which halves the time required to run the test suite (from 53 seconds to 26 seconds) on this machine.

It would not have been possible to reduce this polling period before without significantly augmenting the chances of having multiple test suites stomping on each other's data.

### Deploy Notes
none

### Remaining Todos
* [x] - Find the remaining cases where files are expected in the absolute `/tmp` (they are no longer delivered there)
* [x] - Chase the 85-odd instances of uses of the now deprecated `Settings$.comet`(& friends) trampolines
* [x] - Review whether the Settings instances currently passed around via implicits should remain implicits or should be turned into explicit values (**RFC**)

### DIFF
--- a/src/test/resources/quickstart/metadata/domains/hr.yml
+++ b/src/test/resources/quickstart/metadata/domains/hr.yml
@@ -1,6 +1,6 @@
 name: "hr"
-directory: "/tmp/incoming/hr"
-metadata:
+directory: "£COMET_TEST_ROOT£/incoming/hr"
+metadata:>
   mode: "FILE"
   format: "JSON"
 schemas:
--- a/src/test/resources/quickstart/metadata/domains/hr.yml
+++ b/src/test/resources/quickstart/metadata/domains/hr.yml
@@ -1,5 +1,5 @@
 name: "hr"
-directory: "£COMET_TEST_ROOT£/incoming/hr"
+directory: "__COMET_TEST_ROOT__/incoming/hr"
 metadata:>
   mode: "FILE"
   format: "JSON"
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,6 @@ src/test/resources/application.conf
 # Unix traditional
 *~
 ~*
+
+#Mac
+.DS_Store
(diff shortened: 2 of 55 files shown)


cchepelov:
56 → 29 seconds per test run on an older machine

cchepelov:
Still not too happy with my peppering of `implicit settings: Settings`stuff. WDYT @hayssams ?

src/test/resources/quickstart/metadata/domains/hr.yml
```diff
@@ -1,6 +1,6 @@
 name: "hr"
-directory: "/tmp/incoming/hr"
-metadata:
+directory: "£COMET_TEST_ROOT£/incoming/hr"
```
hayssams: let's use a chair available on every keyboard layout

cchepelov: done. Using __ now.

cchepelov:
Closes: #95

Issue #103, opened 2020-02-28
[BUG] - MatchError in Attribute.checkValidity()


mhdkassir:
## Description
An `Attribute` can have a default value only when it is not obligatory (required = false). This behavior is verified by the `Attribute.checkValidity` method. A MatchError at the line 109 in Attribute.scala prevents this method from executing correctly when the default attribute is set.

## Expected behavior
When the `Attribute.default`property is set, the `Attribute.checkValidity` returns 
- Right(true) if the attribute is not required
- Left (List("..some_error_msg..")) if the attribute is required

## Current Behavior
A MatchError is raised at the line 109 in Attribute.scala when the checkValidity method is invoked on an Attribute instance that has its default value set.
This error can be observed when running the watch step.

## Steps to reproduce
1. Define a Schema (yml) where one of its attributes is required (required: true) and has a default value (default: "some_value")
2. Run the watch step
3. Observer a MatchError in the output.

## Possible implementation
Add a Try on line 109 to match correctly with Success and Failure possibilities

## Complexity estimation
T-Shirt Size: XS

mhdkassir:
correction suggested in PR #104

mhdkassir:
issue close after merging PR #104

PR #104, opened 2020-02-28
Bug fix/match error

mhdkassir:
## Summary
A correction for the bug described in the issue 103

**Related Issue:  #103**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
Enclose the call of Type.sparkValue in a Try block

### How has this been tested?
Unit Test has been added.

### Deploy Notes
N/A


src/main/scala/com/ebiznext/comet/schema/model/Attribute.scala
```diff
@@ -105,8 +105,8 @@ case class Attribute(
       primitiveType.foreach { primitiveType =>
         if (primitiveType == PrimitiveType.struct)
           errorList += s"attribute with name $name: default value not valid for struct type fields"
-        tpe.foreach {
-          _.sparkValue(default) match {
+        tpe.foreach { someTpe =>
+          Try(someTpe.sparkValue(default)) match {
```
cchepelov: Good catch @mhdkassir! Indeed sparkValue is returned as an Any but is actually much more restricted in the actual returned type

I'm not sure there is much value in spinning up a Try .. pattern match on the result vs. a plain `try { } catch { case _: Throwable =>   .... }` block, but this looks good

mhdkassir: right, a try {} catch {} would have been enough. I was focused on fixing the error and didn't give further look into the value of the following structure

PR #138, opened 2020-03-06
BugFix : Use calculated BQ schema instead of inferred

mhdkassir:
## Summary
Use a calculated BQ Schema base on the comet-schema when loading a dataset to BQ instead of using inferred schema

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change: No**

### Solution
Pass the BQ Schema in the parametrs of BQ Load Job

### How has this been tested?
No regression in existing Unit tests
Running ingestion and bq load jobs on GCP

### Other changes
None

### Deploy Notes
N/A

## Contributor checklist:


mhdkassir:
@hayssams ingestion and bq load jobs are OK on GCP, tables are created with the correct schema

Issue #143, opened 2020-03-09
[FEATURE] - Support delivering downstream data via JDBC


cchepelov:
(as a continuation of #96 — should deliver downstream data too)

hayssams:
Done

Issue #136, opened 2020-03-04
[WIP-FEATURE] - Run Autojob against multiple backend (JDBC / BQ / Parquet)


hayssams:
## Description
Currently an auto job executes SQL requests against tables backed by a parquet storage.
This feature intend to add tables backed by other storages including JDBC & BQ.
## Additional context
This could include:
- Design diagrams
- Documentation draft
- Migration strategy
- Adoption and teachability for this feature

# Life cycle

## Possible implementation
Describe here a brief for any contributor wishing to work on this issue

## Complexity estimation
T-Shirt Size: XS, S, M, L. If you would recommend XL or more, we advise you to discuss with the community how to split this issue.
Don't forget to tag this issue as a 'good first issue' if you think it is appropriate!

cchepelov:
We could have a `CompositeSink` that takes a list of sinks, and passes the data on to each successively.

One question: what if one of the sub-sinks fails; should we fail all? should we fail one? Should the ingestion itself be considered failed and to be redone? What then, if the first non-failed sub-sink is not idempotent, how do we deal with duplicates while we re-attempt the failed sub-sinks and the non-tried sub-sinks? 

Another option would be to *refuse* to have `CompositeSink` and instead decide the way to go is to have a single `MessageQueueSink`(e.g. `KafkaSink`, `PulsarSink`, whatever) and then as many MQSource→IndexSink consumers. 

Whichever side the coin lands, perhaps we should write an ADR for this?

mhdkassir:
My understanding of the description of this feature is that _autojobs_ should be able to query data in formats/storage different than parquet files, but @cchepelov question seems to be rather about how to store the result of an _autojob_ in different sinks. @hayssams , the feature concerns the the input data of an _autojob_ or its result storage ?

hayssams:
@mhdkassir @cchepelov 
The composite sink is interesting as a starter.
Actually we need to have a generic reader and writer to support the following flow 
Reader -> Autojob  -> Writer where readers & writers are able to respectively read and writer parquet / jdbc/ BQ sources & sinks.
I will draft a simple solution and share it soon  to get your feedbacks.

cchepelov:
I like this

[...]

PR #148, opened 2020-03-11
[WIP] [CDP-143] Business jdbc delivery

cchepelov:
## Summary
This PR implements JDBC delivery of Business Area data (under the "index" moniker pending #136 

**Related Issue: #143**

**PR Type: Feature**

**Status: WIP**

**Breaking change? Yes**
There are slight changes in the configuration of the Metadata structure as it stands in the Settings (HOCON) and in the Domain definition (YAML) in order to accomodate for JDBC Delivery.

These changes **may break established non-JDBC deliveries** (but won't do so silently and will provide clues/advice on how to adjust)

The most breaking change is that the Metadata parser will no longer accept unknown properties *(we can discuss whether to accept certain or all such unknown properties if wanted, at a policy level)*

### Solution
- implement "Index Sink" delivery via JDBC in the IngestionWorkflow
- flatten any non-scalar fields into JSON, as hardly any database supports non-scalars and those that can tend to support JSON natively anyway.

### How has this been tested?
- Unit Tests currently succesfully log an insert *but* the test should be completed to validate the contents

### Other changes
- merge IndexSinkSettings and IndexSink (that *was* redundant)
- remove custom Jackson parsers for Metadata and Partition (telling Jackson what to do is sufficient, leveraging Jackson's extensive schema validation and error handling)
- ensure exceptions at Index time are not swallowed silently 

### Deploy Notes
- The running `application.conf` should describe the tables into which to perform the data delivery, and should likely provide the DDL to do so. These tables are (concisely) referenced by the `domain.yml` file.
- The point above should be in the documentation
- There may be further refactoring of the configuration to do, in order to split the JDBC Tables definition from the actual application config and domain definition ? Could also have implications on BigQuery and/or Kafka+Avro ?

### Remaining Todos
* [ ] - Update the documentation, possibly including migration notes from 0.12 
* [ ] - Complete the Unit Tests to at least check what's being delivered.
* [ ] - *(nice to have)* a strong consideration to refactor all cases of IndexSink creation (wherever we match on the IndexSink ADT) into a single primitive that can be used wherever we need to Sink something (index, audit, rejected, *business* so far). This exercise of DRY may slightly impact again the documentation but reduce the code at constant service level.

## Contributor checklist:
- [x] My change requires a change to the documentation. ***(very likely)*** 


cchepelov:
@mhdkassir you may be impacted by the config format changes (especially if you sink to BigQuery). Please check whether I wasn't *too* enthusiastic about breaking things 😅

src/main/scala/com/ebiznext/comet/config/IndexSink.scala
```diff
@@ -0,0 +1,77 @@
...
+  putting the annotations up there would cause. Fortunately, Jackson allows us a way out, the price is to never forget
+  to register this "module":
+   */
+  lazy val JacksonModule: SimpleModule = new SimpleModule()
+    .setMixInAnnotation(classOf[IndexSink], classOf[JacksonDummy])
+
+  /**
+    * A no-operation Index Output (disabling external output beyond the business area parquets)
+    */
+//   @JsonDeserialize(builder = classOf[None.NoneBuilder])
```
cchepelov: I've gotta clean this up

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -153,42 +157,32 @@ trait IngestionJob extends SparkJob {
...
-          res match {
-            case Success(_) => ;
-            case Failure(e) => logger.error("JDBCLoad Failed", e)
-          }
-        }
+        val jdbcConfig = JdbcLoadConfig.fromComet(
+          jdbcName,
+          settings.comet,
+          Right(mergedDF),
+          outputTable = schema.name,
```
cchepelov: In the BigQuery case, since we had a "dataset" field in the config, I've used it to allow overriding the destination table. Probably ought to do it "here" too (but actually, probably after factoring out this whole patmat thing)

src/main/scala/com/ebiznext/comet/job/jdbcload/JdbcLoadJob.scala
```diff
@@ -49,24 +53,40 @@ class JdbcLoadJob(
...
     }
+  }
+
+  /**
+    * Just to force any spark job to implement its entry point using within the "run" method
+    *
+    * @return : Spark Session used for the job
+    */
+  override def run(): Try[SparkSession] = {
+    val res = runJDBC()
```
cchepelov: Something significant happens here. Before, we had a double level of cheks. First we had the "master patmat" on the IndexSink ADT *(in today's terms)*, then we checked whether Audit was active if we were on the Audit branch, and then once running the job, we checked again whether Audit was active and delivery to JDBC was requested.

This no longer works in the case where we want JDBC Delivery of Business data but don't care about JDBC delivery of Audit data. 

I think that the sink jobs should not check again (in MetricsJob that is), and just trust that they've been invoked for good reason. 

WDYT?

src/main/scala/com/ebiznext/comet/config/IndexSink.scala
```diff
@@ -0,0 +1,77 @@
...
+/**
+  * Configuration for [[IndexSink]]
+  *
+  * This is used to define an auxiliary output for Audit or Metrics data, in addition to the Parquets
+  * The default Index Sink is None, but additional types exists (such as BigQuery or Jdbc)
+  *
+  */
+sealed abstract class IndexSink( /*val*/ `type`: String) {}
+
+object IndexSink {
```
cchepelov: Née IndexSinkSettings, but also absorbing the role of another IndexSink enum.

In the end, I wonder whether this shouldn't return as IndexSinkSettings, while a single IndexSink object carries the "how do we spin up an IndexSink delivery" ?

mhdkassir:
> @mhdkassir you may be impacted by the config format changes (especially if you sink to BigQuery). Please check whether I wasn't _too_ enthusiastic about breaking things 😅

@cchepelov thanks for drawing my attention, I will take a thorough look and make a feedback

cchepelov:
(of course there is something odd with Scala 2.12 now, thank you Travis. `TODO ++= _` )

cchepelov:
(rebased on current master)

hayssams:
solved by #363

PR #152, opened 2020-03-16
Bugfix/metric bq support

AmineSagaama:
## Summary
The purpose of this PR is to modify the interface contract of Metrics, in order to be readable by 
BigQuery
**Related Issue: #149**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

## The purpose of this PR is to modify the interface contract of Metrics, in order to be readable by 
BigQuery
### Solution
Change the structure of Metrics Object, now, we have an array of struct, instead of simple fields.

### How has this been tested?
You can, trigger an ingestion with metrics set to true, after that, you can export the parquet output files into a Table in BigQuery, and after that, query the table.

### Other changes
Describe any other additional work done:
bug fixes, tweaks, small refactors

### Deploy Notes

### Remaining Todos
* [ ] - Task1
* [ ] - Task2

## Contributor checklist:
- [ ] My code follows the code style of this project.


cchepelov:
the JDBC section LGTM 👍

AmineSagaama:
@hayssams Could you check this PR ?

src/main/resources/reference.conf
```diff
@@ -166,7 +166,7 @@ jdbc-engines {
         percentile25 INTEGER NULL,
         median INTEGER NULL,
         percentile75 INTEGER NULL,
-        catCountFreq VARCHAR(255) not NULL,
+        catCountFreq LONGTEXT not NULL,
```
cchepelov: not sure we need to go to a CLOB on this (we'll need to load everything in memory for JSON rehydratation anyway). Probably good enough to stick with plain unlimited VARCHAR

src/main/resources/reference.conf
```diff
@@ -166,7 +166,7 @@ jdbc-engines {
...
         countDistinct INTEGER NULL,
         missingValuesDiscrete INTEGER NULL,
         count INTEGER not NULL,
```
cchepelov: The Postgres side of the schema likely requires some updating as well

AmineSagaama: yes 👍

src/main/resources/reference.conf
AmineSagaama: use TEXT type is enough so ?

src/main/resources/reference.conf
```diff
@@ -228,6 +228,7 @@ jdbc-engines {
         percentile25 INTEGER NULL,
         median INTEGER NULL,
         percentile75 INTEGER NULL,
+        catCountFreq LONGTEXT not NULL,
```
cchepelov: Postgres lacks LONGTEXT. Please use TEXT instead

src/main/resources/reference.conf
cchepelov: TEXT is an alias for CLOB just as LONGTEXT on H2; VARCHAR is the one that remains a string
(on the other hand, and more accurately, second thought... We don't need to have this field be 'stored' in-heap on the H2 side of the JDBC link, since we are not going to address any data within. So your initial move *was* correct, we can store that as CLOB on H2; "TEXT" would be appropriate.

On *Postgres* this is a different story: Postgres *can* do meaningful operations on JSON-stored-as-text (we can e.g. `select * from metrics where catCountFreq:json @> 'some_json_expression'), so the proper type would be VARCHAR (without limit, a Postgres extension) or TEXT (also a Postgres extension, an alias to VARCHAR)

cchepelov:
Thank you @AmineSagaama!

PR #155, opened 2020-03-19
Domain to database export scripts

mmenestret:
## Summary

Generating one or more export scripts, based on a Mustache template and an Excel referential to export database tables.

**Related Issue: #IssueId**

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### Solution

Generates extraction scripts based on a given domain
The schemas should at least, specify :
    - a table name (schemas.name)
    - a file pattern (schemas.pattern) which is used as the export file base name
    - a write mode (schemas.metadata.write): APPEND or OVERWRITE
    - a delta column (schemas.merge.timestamp) if in APPEND mode : the column which is used to determine new rows for each exports
    - the columns to extract (schemas.attributes.name*)

In the Mustache (http://mustache.github.io/mustache.5.html) template file, you'll write your extraction export process (sqlplus for Oracle, pgsql for PostgreSQL as an example).

In that template you can use the following parameters:

- table_name  -> the table to export
- delimiter   -> the resulting dsv file delimiter
- columns     -> the columns to export
   columns is a Mustache map, it gives you access, for each column, to:
    - name               -> the column name
    - trailing_col_char  -> the separator to append to the column (, if the is more columns to come, "" otherwise)
    Here is an example how to use it in a template:
    SELECT
    {{#columns}}
    TO_CHAR({{name}}){{trailing_col_char}}
    {{/columns}}
    FROM
    {{table_name}};
- export_file -> the export file name
- full_export -> if the export is a full or delta export (the logic is to be implemented in your script)

### How has this been tested?

Unit tests are done to verify the generation of the template parameters and also the generation of a script based on a given Mustache template and a set of template parameters.

## Contributor checklist:
- [ X ] My code follows the code style of this project.
- [  ] My change requires a change to the documentation.
- [ X ] I have updated the documentation accordingly.


src/main/scala/com/ebiznext/comet/database/extractor/TemplateParams.scala
```diff
@@ -0,0 +1,99 @@
...
+      .map(_.toUpperCase)
+      .foldLeft(
+        List(
+          "table_name"  -> tableToExport.toUpperCase,
+          "delimiter"   -> dsvDelimiter,
+          "columns"     -> columnsParam,
+          "export_file" -> exportOutputFileBase,
+          "full_export" -> fullExport
+        )
+      ) { case (list, deltaCol) => list :+ ("delta_column" -> deltaCol.toUpperCase) }
```
mhdkassir: `toUpperCase` seems to have already been applied at line 44

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+    * Generate all extraction scripts based on the given domain
+    * @param domain The domain extracted from the Excel referential file
+    * @param scriptTemplateFile The script template
+    * @param scriptsOutputPath Where the scripts are produced
+    * @return The list of produced files
+    */
+  def generate(domain: Domain, scriptTemplateFile: File, scriptsOutputPath: File): List[File] = {
+    val preEncryptionDomain = SchemaGen.genPreEncryptionDomain(domain)
```
mhdkassir: I think no need to invoke `genPreEncryptionDomain`, you can use the `domain` parameter directly

src/test/scala/com/ebiznext/comet/schema/generator/SchemaGenSpec.scala
```diff
@@ -17,12 +17,14 @@ class SchemaGenSpec extends TestHelper {
       val schema1 = result.schemas.filter(_.name == "SCHEMA1").head
       schema1.metadata.flatMap(_.format) shouldBe Some(Format.POSITION)
       schema1.attributes.size shouldBe 19
+      schema1.merge.flatMap(_.timestamp) shouldBe Some("UPDATE")
```
mhdkassir: perhaps it would be better to choose one of the existing columns of SCHEMA1 as the timestamp column. "UPDATE" is not a real column. Or add a row for UPDATE in the SCHEMA1 tab

src/main/scala/com/ebiznext/comet/database/extractor/TemplateParams.scala
mmenestret: You're right, fixed.

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
mmenestret: Ok fixed !

src/test/scala/com/ebiznext/comet/schema/generator/SchemaGenSpec.scala
mmenestret: Ok fixed !

mmenestret:
I have a suggestion about my PR.

Today, the main expects an Excel file.

I suggest to decouple it from the Excel file and make it an optional param.
- If present, it changes nothing
- If not, it expects to find yaml files just like for a regular ingestion

What's your take on this ?

mhdkassir:
> I have a suggestion about my PR.
> 
> Today, the main expects an Excel file.
> 
> I suggest to decouple it from the Excel file and make it an optional param.
> 
>     * If present, it changes nothing
> 
>     * If not, it expects to find yaml files just like for a regular ingestion
> 
> 
> What's your take on this ?

Generating sql scripts from yaml files directly can be useful but not passing any parameter can be a source of confusion. I suppose you will look for yaml files in metadata/domain , what if there are more than one yaml (domain) files. Will the program generate scripts for all of them ? Perhaps we could pass the domain name as a parameter that can be used in both scenarios (excel and yaml)

mmenestret:
Yes, or at least make it mandatory in case of yaml file based generation and optional in excel file based generation (since it won't be used).
I re-open the PR and we'll do it as soon as I can.

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
```diff
@@ -0,0 +1,117 @@
...
+  *
+  * Here you'll write your extraction export process (sqlplus for Oracle, pgsql for PostgreSQL as an example).
+  * In that template you can use the following parameters:
+  *
+  * table_name  -> the table to export
+  * delimiter   -> the resulting dsv file delimiter
+  * columns     -> the columns to export
+  * columns is a Mustache map, it gives you access, for each column, to:
+  *  - name               -> the column name
+  *  - trailing_col_char  -> the separator to append to the column (, if the is more columns to come, "" otherwise)
```
AmineSagaama: "if there are more columns to come"

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
```diff
@@ -0,0 +1,117 @@
...
+object Main extends App with StrictLogging {
+
+  import ScriptGen._
+
+  val arglist = args.toList
+
+  logger.info(s"Running Comet $arglist")
+
+  ExtractScriptGenConfig.parse(args) match {
+    case Some(config) =>
```
AmineSagaama: I propose to add a case class here, in order to check if it's an excel configuration or something else, therefore, we will have a case for each config.

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGenConfig.scala
```diff
@@ -0,0 +1,84 @@
...
+  scriptTemplateFile: File = File("."),
+  scriptOutputDir: File = File(".")
+)
+
+object ExtractScriptGenConfig {
+
+  val builder = OParser.builder[ExtractScriptGenConfig]
+
+  def exists(name: String)(path: String): Either[String, Unit] =
+    if (File(path).exists) Right(())
```
AmineSagaama: maybe, we should return a right of Boolean here instead of Unit, what do you think ?

src/main/scala/com/ebiznext/comet/database/extractor/TemplateParams.scala
```diff
@@ -0,0 +1,98 @@
...
+  * @param columnsToExport cols to export
+  * @param fullExport if table is going to be fully or delta exported
+  * @param deltaColumn if delta exported, which is the col holding the date of last update
+  * @param dsvDelimiter export result dsv delimiter
+  * @param exportOutputFileBase export dsv file base name (will be completed by current datetime when executed)
+  * @param scriptOutputFile where the script is produced
+  */
+case class TemplateParams(
+  tableToExport: String,
+  columnsToExport: List[String],
```
AmineSagaama: if we choose fullExport as option, maybe we should convert columnsToExport to Option[List[String]] ?

src/main/scala/com/ebiznext/comet/database/extractor/TemplateParams.scala
```diff
@@ -0,0 +1,98 @@
...
+  fullExport: Boolean,
```
AmineSagaama: what do you think of putting it false by default ?

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
mmenestret: I finally turned it into a yml to scripts functionality only.
If needed, users have to generate yml from Excel before.

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
mmenestret: Fixed

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGenConfig.scala
mmenestret: Scopt expects something with that form: `def validate(f: A => Either[String, Unit]): OParser[A, C]`.
Anyway, since if file does not exist we would have to return a `Left`, we would have always had to return a `Right(true)` if the file exists, a `Right(false)` would never be used.
Thus it is isomorphic to Right(()). Do you see my point ?

src/main/scala/com/ebiznext/comet/database/extractor/TemplateParams.scala
mmenestret: By full export, every single rows, not every columns. If one needs to export every columns anyway, one can modify its mustache template to select * and not to use the columns template params.

src/main/scala/com/ebiznext/comet/database/extractor/TemplateParams.scala
mmenestret: It is a decision to make, but the delta export has more constraints (needs a delta column specified).
It already defaults as full export there: `val isFullExport = schema.metadata.flatMap(_.write).contains(WriteMode.OVERWRITE)`

src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGenConfig.scala
AmineSagaama: yes

Issue #160, opened 2020-03-29
[BUG] - Pattern compiled for every attribute of every row


hayssams:
## Description
Comet check for pattern on each attribute in each row. We currently compile the pattern on every attribute of every row which lead to a huge performance hit.
## Expected behavior
Since the Pattern class is thread safe, w only need to compile the pattern just once and instantiates the matcher on each attribute.
Pattern should be compiled on schema load.

## Current Behavior
Pattern compiled during ingestion 
Affects performance

## Complexity estimation
XS

PR #161, opened 2020-03-29
Pattern compile

hayssams:
## Summary
See #160 

Merge PR #159 first

### How has this been tested?
Covered by the existing TypeSpec Unit Test


project/plugins.sbt
```diff
@@ -26,4 +26,6 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.8")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.1-M3")
 
 // provides "sbt dependencyUpdates":
-addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.5.0")
\ No newline at end of file
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.5.0")
+
+addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
```
cchepelov: It looks like we have another feature freeriding this PR here..

src/main/scala/com/ebiznext/comet/job/Main.scala
```diff
@@ -122,25 +126,29 @@ object Main extends StrictLogging {
...
-        val schema = arglist(2)
-        val paths = arglist(3)
-        val lockPath = new Path(settings.comet.lock.path, s"${domain}_${schema}.lock")
-        val locker = new FileLock(lockPath, storageHandler)
-        val waitTimeMillis = settings.comet.lock.ingestionTimeout
-
-        locker.doExclusively(waitTimeMillis) {
-          workflow.ingest(domain, schema, paths.split(',').map(new Path(_)).toList)
-        }
+      case "ingest" =>
```
cchepelov: Doesn't scopt already have a "command" / "major mode" facility? why not use it?

(would this PR benefit from rebasing, or do we have a second freerider here?)

src/main/scala/com/ebiznext/comet/job/Main.scala
```diff
@@ -122,25 +126,29 @@ object Main extends StrictLogging {
...
+        IngestConfig.parse(args.drop(1)) match {
```
cchepelov: args.tail

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
```diff
@@ -69,6 +69,39 @@ case class Type(
...
+      case "string" => None
+      case _ =>
+        primitiveType match {
+          case PrimitiveType.struct => None
+          case PrimitiveType.date =>
+            None
+          case PrimitiveType.timestamp =>
+            None
+          case PrimitiveType.boolean =>
+            val tf = pattern.split("<-TF->")
```
cchepelov: Java String's `.split`method itself uses regexes
Would suggest using .indexOf and manually cutting the string instead (for safety).

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
```diff
@@ -80,9 +113,11 @@ case class Type(
           case PrimitiveType.timestamp =>
             Try(timestamp.fromString(value, pattern, zone.getOrElse(null))).isSuccess
           case PrimitiveType.boolean =>
-            boolean.matches(value, pattern)
+            // We can get the pattern safely since checkValidity has been called by now
+            boolean.matches(value, compiledPattern.get.right.get._1, compiledPattern.get.right.get._2)
```
cchepelov: can we find a way to avoid this `.get.right.get._1` business? It looks like we are going to do a lot of effectful, potentially exception-throwing complex path dereference, which might be optimized / elided

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
```diff
@@ -80,9 +113,11 @@ case class Type(
...
           case _ =>
-            val textPattern = Pattern.compile(pattern, Pattern.MULTILINE)
+            // We can get the pattern safely since checkValidity has been called by now
+            val textPattern = compiledPattern.get.left.get
```
cchepelov: ditto

src/main/scala/com/ebiznext/comet/schema/model/WriteMode.scala
```diff
@@ -51,13 +51,14 @@ sealed case class WriteMode(value: String) {
...
 object WriteMode {
-
   def fromString(value: String): WriteMode = {
     value.toUpperCase() match {
       case "OVERWRITE"       => WriteMode.OVERWRITE
       case "APPEND"          => WriteMode.APPEND
       case "ERROR_IF_EXISTS" => WriteMode.ERROR_IF_EXISTS
       case "IGNORE"          => WriteMode.IGNORE
+      case _ =>
+        throw new Exception(s"Invalid Write Mode try one of ${writes}")
```
cchepelov: (not specifically here) wondering whether we shouldn't transition these enums to enumeratum, which already provides these services and prepares nicely for dotty?

hayssams: I am opening an specific issue to address it

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
hayssams: Done

project/plugins.sbt
hayssams: This PR is based on  #159 which should be merged first

cchepelov: Indeed. I seem to have seen @mhdkassir reviewing that one, yielding

src/main/scala/com/ebiznext/comet/job/Main.scala
cchepelov: (deleted "freerider" comment — that's actually a serial dependency on #159)

cchepelov:
LGTM (pending #159)

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
hayssams: done

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
hayssams: done

PR #162, opened 2020-03-29
Add encryption algorithms

hayssams:
## Summary
Very early stage of encryption algorithms allowing us have a real data masking feature.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### Solution
Describe your changes in detail

### How has this been tested?
Outline the steps taken to test the PR (unit tests? Integration tests? tested impacts on other project features? ...)

### Other changes
Describe any other additional work done:
bug fixes, tweaks, small refactors

### Deploy Notes

### Remaining Todos
* [ ] - Task1
* [ ] - Task2

## Contributor checklist:
- [ ] My code follows the code style of this project.


cchepelov:
Just to make sure, can we clarify whether the "encryption" name here is apt?

Is it about 

- reversible encryption (to ensure safety of data at rest and/or provide opaque data delivery)
- data hashing (to destroy or at least fuzz PII while keeping aggregate statistics) ?

hayssams:
@cchepelov We currently only support data masking. 
The reversible AES is not yet implmented. This package is supposed to handle both.
We probably should move masking functions to a Masking package.

cchepelov:
Indeed if this is about data masking, not reversible encryption, I think we ought to call it that way.

Do we have semantic tagging at this point ?


(On the other hand, if one day we support:
- at-rest encryption using one of the standard reversible algorithms (AES, GOST R 34.12-2015 "Kuznyechik", Blowfish etc.)
- opaque statistics via holomorphic encryption algos
this might be awesome)

hayssams:
> Indeed if this is about data masking, not reversible encryption, I think we ought to call it that way.
> 
> Do we have semantic tagging at this point ?
> 
> (On the other hand, if one day we support:
> 
> * at-rest encryption using one of the standard reversible algorithms (AES, GOST R 34.12-2015 "Kuznyechik", Blowfish etc.)
> * opaque statistics via holomorphic encryption algos
>   this might be awesome)



> this assert is now guaranteed to never trigger

removed

hayssams:
> not really sure about `def encrypt(value: String)` -> `def crypt(value: String)` (from a purely linguistic POV)
> 
> What about just `def apply(value: String)` ?

The word Encryption covers data masking and encryption. 
The linux command line "crypt" function based on DES algorithm is actually a data masking feature.

hayssams:
> not really sure about `def encrypt(value: String)` -> `def crypt(value: String)` (from a purely linguistic POV)
> 
> What about just `def apply(value: String)` ?

For reversible encryption, we will probably have a decrypt method. Not sure yet, we may leave it like that and change it later if decrypt does not make it later.

cchepelov:
I agree. Let's keep "encrypt" then? + maybe a comment in the trait, such as

"@note the strength guarantees and/or reversibility properties of the encryption are not specified at this time"

cchepelov:
> The linux command line "crypt" function based on DES algorithm is actually a data masking feature.

True. Although early UNIX command & function names were frequently truncated due to purely 70's dinosaur constraints, not linguistic or correctness questions... The choice of that function's name has been the subject of debate and lament in the cryptographer community

PR #173, opened 2020-04-01
Handle correctly subdirectories in Parquet2CSV

hayssams:
## Summary
When source domain and/or schema name is not specified, parquet2csv does not handle correctly subdirectory browsing and thus does not convert any parquet file.

This PR also add an option to delete the source parquet file after completion (false by default)

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/convert/Parquet2CSVConfig.scala
+++ b/src/main/scala/com/ebiznext/comet/job/convert/Parquet2CSVConfig.scala
@@ -11,10 +11,9 @@ case class Parquet2CSVConfig(
   outputFolder: Option[Path] = None,
   domainName: Option[String] = None,
   schemaName: Option[String] = None,
-  withHeader: Boolean = false,
   writeMode: Option[WriteMode] = None,
   deleteSource: Boolean = false,
-  separator: String = ",",
+  options: List[(String, String)] = Nil,
   partitions: Int = 1
 )
 
@@ -26,7 +25,7 @@ object Parquet2CSVConfig extends CliConfig[Parquet2CSVConfig] {
     OParser.sequence(
       programName("comet"),
       note(
-        "example => --input_dir /tmp/datasets/accepted/ --output_dir /tmp/datasets/csv/ --domain sales --schema orders --with_header  --separator ,  --partitions 1 --write_mode overwrite"
+        "example => --input_dir /tmp/datasets/accepted/ --output_dir /tmp/datasets/csv/ --domain sales --schema orders --option header=true  --option separator=,  --partitions 1 --write_mode overwrite"
       ),
       head("comet", BuildInfo.version),
       opt[String]("input_dir")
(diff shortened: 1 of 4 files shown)


src/main/scala/com/ebiznext/comet/job/convert/Parquet2CSVConfig.scala
```diff
@@ -13,6 +13,7 @@ case class Parquet2CSVConfig(
   schemaName: Option[String] = None,
   withHeader: Boolean = false,
   writeMode: Option[WriteMode] = None,
+  deleteSource: Boolean = false,
   separator: String = ",",
   partitions: Int = 1
```
mmenestret, line 18: About Config, shouln't we take the schema parameters if they are not overriden by command line params ?
And maybe we should also add a "quoted" param to signify if the values are surrounded by quotes ?

cchepelov: Re. your first point, it makes sense. It can be handled by having a Parquet2CSVConfig.apply(schemaparameters) and using it to provide the initial instance to the scopt "mas-o-mesos-monad"

regarding your second point, do you mean a way to request a deviation from regular practice RFC 4180 where every data field is quoted (rather than only when needed)? (to be clear, this is not a deviation from RFC 4180 per se, clause 2.5 effectively supports systematic quoting if sender so chooses)

mmenestret: 1. Yes but the way it is implemented for now, it uses the same Config for all parquet to csv transformations since the dependency is injected during class instantiation and not during `run()` calls.
2. At the moment, the generated CSV do not quote their fields, so it is more a request to be closer to the RFC (which I did not know about before your reply !)

cchepelov: 1. ugh, good catch

2. what do you mean ? If a value is the contents of 
```
val foo = "abc,def\"ghi\"jkl,mno"
```
do you mean that it will spit out 

> _previousfield_,abc,def"ghi"jkl,mno,_nextfield_

rather than 

> _previousfield_,"abc,def""ghi""jkl,mno",_nextfield_

as it ought to?

regarding RFC 4180: the fun thing is it effectively codifies the maze of things done by Microsoft Excel and others, decades after the fact ;)

mmenestret: About point 1, I'll create an issue and probably a PR for it.
About point 2. I think I got it wrong, the behavior of spark-csv depends on
`quoteMode: when to quote fields (ALL, MINIMAL (default), NON_NUMERIC, NONE), see Quote Modes` so it is up to the user to set the behavior he wants.

cchepelov: ... point 2, indeed Spark is standards-compliant on that regard ;)

Issue #175, opened 2020-04-04
[BUG] - options should be unbounded in Parquet2CSV


hayssams:
## Description
See https://github.com/ebiznext/comet-data-pipeline/issues/174#issuecomment-609002264

## Complexity estimation
T-Shirt Size: XS

PR #176, opened 2020-04-04
Allow the option argument to appear multiple times in parquet2csv

hayssams:
## Summary
option argument in parquet2cscv is allowed to appear multiple times.
**Related Issue: #IssueId**
#175 
**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


src/main/scala/com/ebiznext/comet/job/Main.scala
```diff
@@ -195,7 +195,7 @@ object Main extends StrictLogging {
           case Some(config) =>
             new Parquet2CSV(config, storageHandler).run()
           case _ =>
-            println(MetricsConfig.usage())
+            println(Parquet2CSVConfig.usage())
```
hayssams, line 198: Well spotted

PR #190, opened 2020-04-09
Indexing in ES / BQ / JDBC should be active if active.index is not valued to None

hayssams:
## Summary
Audit may be active or not. In this PR, audit is always active. Only indexing is optional.
Also, Indexing was not effective when audit.active is false which make no sense.
To disable indexing, simply set comet.audit.index to None
**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -57,10 +57,6 @@ hadoop {
 }
 
 audit {
-  active = true
-  active = ${?COMET_AUDIT_ACTIVE}
-
-  #  path = ${root}"/metrics/{domain}/{schema}"
   path = ${root}"/audit"
   path = ${?COMET_AUDIT_PATH}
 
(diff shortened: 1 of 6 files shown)


src/main/resources/reference.conf
```diff
@@ -60,10 +60,6 @@ hadoop {
 }
 
 audit {
-  active = true
```
hayssams, line 63: Audit is now always active

mmenestret: Maybe sometime we still want to switch audit off don't you think ?

hayssams: You're right. But we always logs in the rejected folder the faulty records.
audit adds up in the audit folder global statistics about the ingestion process. Let's keep it like that for now and think about making it active later

Issue #217, opened 2020-04-28
[BUG] - BigQuery can't infer the Schema of the Dataset if the Table is based on partitioning


AmineSagaama:
## Description
When we want to partition a Table in BigQuery, we have to explicitly specify the schema of the Dataset, otherwise we will have such an error : 
BigQuery error in load operation: Table with field based partitioning must have a schema.'

## Expected behavior
BigQuery can import a partitioned table without any errors.

## Current Behavior
An error is raised when we try to import a **Partitioned** Table from an existing dataset (cloud storage)

## Possible implementation
We have to specify the schema of the Table before importing it.

## Complexity estimation
T-Medium Size: M

mhdkassir:
@AmineSagaama I think this issue can be closed

AmineSagaama:
fixed here #219

PR #219, opened 2020-04-29
[FIX] BigQuery can't infer the Schema of the Dataset if the Table is Partitioned

AmineSagaama:
## Summary
A correction to the problem described in #217 

**Related Issue: #217**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
We have to pass the schema to bq for **Partitioned** Table.

### How has this been tested?
Run an autojob with index BQ active, a Partitioned Table is created in BigQuery, for example :

```
name: "person_agg"
views:
  person_View: "accepted/person/person"
tasks:
  - domain: "person_agg_domain"
    area: "business"
    dataset: "person_agg_kpis"
    write: "OVERWRITE"
    sql: |
      select name, to_date({{date_event}}, "yyyy-dd-MM") AS event_date,
      current_timestamp() AS processing_time 
      from person_View
    partition: ["event_date"]
    index: "BQ"
    properties:
      timestamp: event_date 
```

## Contributor checklist:

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/conversion/Convertible.scala
+++ b/src/main/scala/com/ebiznext/comet/job/conversion/Convertible.scala
@@ -1,0 +1,17 @@
+package com.ebiznext.comet.job.conversion
+
+trait Convertible[A, B] extends (A => B)
+
+trait ConvertibleWith[A, CoA, B] extends ((A, CoA) => B)
+
+package object syntax {
+
+  implicit class ConvertibleOps[A](a: A) {
+
+    def to[B](implicit convert: Convertible[A, B]): B = convert(a)
+
+    def to[B, WithA](adjuvant: WithA)(implicit convert: ConvertibleWith[A, WithA, B]): B =
+      convert(a, adjuvant)
+  }
+
+}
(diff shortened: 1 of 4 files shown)


src/main/scala/com/ebiznext/comet/job/conversion/Convertible.scala
```diff
@@ -0,0 +1,17 @@
+package com.ebiznext.comet.job.conversion
```
hayssams, line 1: too much extra code for limited added value no. Also this include one more implicit :(

src/main/scala/com/ebiznext/comet/job/conversion/bigquery.scala
```diff
@@ -0,0 +1,80 @@
...
+    override def apply(v1: DataFrame): BQSchema = bqSchema(v1)
+  }
+
+  /**
+    *
+    * @param df Spark DataFrame
+    * @return
+    */
+  private[conversion] def bqSchema(df: DataFrame): BQSchema = {
+    def convert(sparkType: DataType): LegacySQLTypeName = {
```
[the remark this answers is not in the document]
AmineSagaama: Agree, I didn't want to change the Schema.scala, but, indeed, we can refactor the code and have on generic function.

src/main/scala/com/ebiznext/comet/job/conversion/Convertible.scala
AmineSagaama: The aim of this class is to have a generic conversion between two DataTypes, now, we need to convert Spark Schema to BigQuery Schema, but in the future, we may need to convert Avro Schema To BigQuery Schema etc..

PR #227, opened 2020-05-04
Schema signed fix

hayssams:
## Summary
A couple of fix :
- remove extra check when handling signed decimals
- Generate YML from XLS in the  metadata/domains/DOMAIN_NAME directory

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


src/main/scala/com/ebiznext/comet/schema/model/PrimitiveType.scala
```diff
@@ -124,9 +124,8 @@ object PrimitiveType {
         val currentLocale: Locale = new Locale(locale(0), locale(1))
         val numberFormatter =
           NumberFormat.getNumberInstance(currentLocale).asInstanceOf[DecimalFormat]
-        str.headOption
-          .withFilter(_.equals('+'))
-          .foreach(_ => numberFormatter.setPositivePrefix("+"))
+        if (str.head == '+')
+            numberFormatter.setPositivePrefix("+"))
```
mhdkassir, line 128: remove extra parenthesis

src/main/scala/com/ebiznext/comet/schema/generator/SchemaGen.scala
```diff
@@ -93,7 +93,7 @@ object SchemaGen extends LazyLogging {
 object Main extends App {
   import SchemaGen._
   implicit val settings: Settings = Settings(ConfigFactory.load())
-  val outputPath = settings.comet.metadata
+  val outputPath = DatasetArea.domains.toString
```
mhdkassir, line 96: Using `DatasetArea` creates a dependency with hadoop libraries, which I think we should  avoid because the yaml generation is done without necessarily running a Spark Job and hence the hadoop libraries (provided) are not available in the Path. Perhaps just build the domains path by concatenating `settings.comet.metadata` with "/domains" ?

hayssams: We use the dependency or we do not.
That would be too difficult to maintain both cases : With or wtihout Spark.
What do you think ?

mhdkassir: definitely with spark

PR #235, opened 2020-05-13
Bugfix/comments in bq schema

mhdkassir:
## Summary
Fix a regression in loading data to BigQuery. The description of the columns of a schema is not present in the BQ tables although specified in the YAML schema

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
In `BigQueryLoadJob`, use the `maybeSchema` when present to create the Table.  

### How has this been tested?
No regression on existing Unit Tests

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/bqload/BigQueryLoadJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/bqload/BigQueryLoadJob.scala
@@ -55,7 +55,7 @@ class BigQueryLoadJob(
           case _ =>
             cliConfig.outputPartition match {
               case Some(_) => StandardTableDefinition.of(df.to[BQSchema]).toBuilder
-              case _       => StandardTableDefinition.newBuilder()
+              case None       => StandardTableDefinition.newBuilder()
             }
         }
 
--- a/src/main/scala/com/ebiznext/comet/job/bqload/BigQueryLoadJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/bqload/BigQueryLoadJob.scala
@@ -52,7 +52,9 @@ class BigQueryLoadJob(
         cliConfig.outputPartition match {
           case Some(_) =>
             StandardTableDefinition.of(df.to[BQSchema]).toBuilder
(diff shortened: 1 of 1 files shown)


src/main/scala/com/ebiznext/comet/job/bqload/BigQueryLoadJob.scala
```diff
@@ -52,7 +52,10 @@ class BigQueryLoadJob(
         cliConfig.outputPartition match {
           case Some(_) =>
             StandardTableDefinition.of(df.to[BQSchema]).toBuilder
-          case _ => StandardTableDefinition.newBuilder()
+          case _ =>
+            maybeSchema.fold(StandardTableDefinition.newBuilder()) { schema =>
+              StandardTableDefinition.of(schema).toBuilder
+            }
```
hayssams, line 58: The differentiator is the presence of the schema
Please correct it as follows : 
```
      val tableDefinitionBuilder =
        maybeSchema match {
          case Some(schema) =>
            StandardTableDefinition.of(schema).toBuilder
          case _ =>
            StandardTableDefinition.of(df.to[BQSchema]).toBuilder
        }

```

mhdkassir: there is also the case where  maybeSchema is None but also  cliConfig.outputPartition is None, where we don't necessarily need to force the schema. What do you think ?
```
      val tableDefinitionBuilder =
        maybeSchema match {
          case Some(schema) =>
            StandardTableDefinition.of(schema).toBuilder
          case _ => cliConfig.outputPartition match{
            case Some(_) => StandardTableDefinition.of(df.to[BQSchema]).toBuilder
            case _ => StandardTableDefinition.newBuilder()
          }
        }
```

hayssams: LGTM

mhdkassir: Done

PR #241, opened 2020-05-15
Fix - Using schema separator and withHeader

mmenestret:
Using schema withHeader and separator value in postEncrypt schema generation instead of default values.


mmenestret:
Fixed test and changing behavior that way:
- If a separator is specified in command line arg, use it
- Otherwise, use schema's separator
- Otherwise, use default value that was used before ("µ")

PR #248, opened 2020-05-20
Feature/autojob on bigquery

hayssams:
## Summary
This PR now allow spark job to load views from big query and push the results directly to a big query table.

To load the views from a bigquery table simply prefix the bigquery table by a BQ: prefix as below
`
views:
  myview1: "BQ:mydataset.mytable" 
  myview2: "BQ:myproject:mydataset.mytable"
`

To store the results in a BQ table choose BQ as an index sink.
To avoid saving the results in a Parquet file on the filesystem, do not define a storage area ini the AutoJob or AutoTask definition.
`tasks:
  \- domain: "customers"
    dataset: "myjob"
    write: "APPEND"
    sql: select * from myview1
    index: BQ
`
In the example below, the results will be saved in the dataset "customers" and the "my job" table.

Spark BigQuery extra properties as described here https://github.com/GoogleCloudDataproc/spark-bigquery-connector are also available by prefixing the keys with  "spark." in the properties section of the AutoTask descriptor

**PR Type:Feature**

**Status: Ready to review**

**Breaking change? Yes**
The Spark BigQuery Client dependency is now declared as provided. Any job running on GCP should now reference the Spark BigQuery Client Lib artefact as explained here     https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/36

## Contributor checklist:
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.


project/Dependencies.scala
```diff
@@ -95,7 +95,10 @@ object Dependencies {
     "com.google.cloud.bigdataoss" % "gcs-connector" % Versions.gcs from gcsConnectorShadedJar exclude ("javax.jms", "jms") exclude ("com.sun.jdmk", "jmxtools") exclude ("com.sun.jmx", "jmxri") excludeAll (jacksonExclusions: _*) classifier ("shaded"),
     "com.google.cloud.bigdataoss" % "bigquery-connector" % Versions.hadoopbq from gcpBigQueryConnectorShadedJar exclude ("javax.jms", "jms") exclude ("com.sun.jdmk", "jmxtools") exclude ("com.sun.jmx", "jmxri") excludeAll (jacksonExclusions: _*) classifier ("shaded"),
     "com.google.cloud" % "google-cloud-bigquery" % Versions.bq exclude ("javax.jms", "jms") exclude ("com.sun.jdmk", "jmxtools") exclude ("com.sun.jmx", "jmxri") excludeAll (jacksonExclusions: _*),
-    "com.google.cloud.spark" %% "spark-bigquery-with-dependencies" % "0.15.1-beta"
+    // see https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/36
+    // Add the jar file to spark dependencies
+    "com.google.cloud.spark" %% "spark-bigquery-with-dependencies" % "0.15.1-beta" % "provided"
```
hayssams, line 100: dependency is now provided. The library should be referenced in the --jars arg of the job in  Composer

src/main/scala/com/ebiznext/comet/config/DatasetArea.scala
```diff
@@ -182,10 +182,6 @@ object StorageArea {
 
   def fromString(value: String)(implicit settings: Settings): StorageArea = {
 
-    val rejected = settings.comet.area.rejected.toLowerCase(Locale.ROOT)
```
hayssams, line 185: We never used them, so let's get rid of it.

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -68,8 +69,27 @@ class AutoTask(
     }
     views.getOrElse(Map()).foreach {
       case (key, value) =>
-        val fullPath = if (value.startsWith("/")) value else s"${settings.comet.datasets}/$value"
-        val df = session.read.parquet(fullPath)
```
hayssams, line 72: We now check if the views have to be loaded from bigquery (BQ Prefix)
Note that parquet files can now be marked with the "FS:" prefix (default)

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -66,10 +67,36 @@ class AutoTask(
           .asInstanceOf[UdfRegistration]
       udfInstance.register(session)
     }
```
hayssams, line 69: properties starting with "spark." are sett in the session conf as described by the spark bigquery connector

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -80,59 +107,65 @@ class AutoTask(
...
-      partitionedDatasetWriter(
-        if (coalesce) dataframe.coalesce(1) else dataframe,
-        task.getPartitions()
-      )
-    partitionedDF
-      .mode(SaveMode.Overwrite)
-      .format(settings.comet.writeFormat)
-      .option("path", mergePath)
-      .save()
+    // Target Path exist only if a storage area has been defined at task or job level
```
hayssams, line 112: target path is now optional. If storage is not defined, not target path is available. The results are directly piped to BQ

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -80,59 +107,65 @@ class AutoTask(
...
+          val csvPath = storageHandler.list(targetPath, ".csv", LocalDateTime.MIN).head
+          val finalCsvPath = new Path(targetPath, targetPath.getName() + ".csv")
+          storageHandler.move(csvPath, finalCsvPath)
+        }
       }
     }
+
     task.postsql.getOrElse(Nil).foreach(session.sql)
-    Success(session)
+    // Lets retturn the Dataframe soo that it can be piped to another sink
```
hayssams, line 168: This is the reason why we return SparkJobResult instead of the session object alone.

src/main/scala/com/ebiznext/comet/schema/model/Types.scala
```diff
@@ -108,22 +111,22 @@ case class Type(
   }
 
   def matches(value: String): Boolean = {
```
hayssams, line 113: This should not be in this PR, but it was done lately. 
To speed up ingestion for striating columns, we evaluated the type of columns early on; Hence the isString.

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -66,10 +67,36 @@ class AutoTask(
           .asInstanceOf[UdfRegistration]
       udfInstance.register(session)
     }
+    task.properties.foreach { properties =>
+      properties.map {
+        case (k, v) if k.startsWith("spark.") =>
```
AmineSagaama, line 72: Could you explain why we have to pass a "spark." properties ? I don't see it in the documentation.

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -66,10 +67,36 @@ class AutoTask(
...
+            (IndexSink.FS, value)
+
+        val df = format match {
+          case FS =>
+            val fullPath =
+              if (path.startsWith("/")) path else s"${settings.comet.datasets}/$path"
+            session.read.parquet(fullPath)
+          case BQ =>
+            session.read
+              .format("com.google.cloud.spark.bigquery")
```
AmineSagaama, line 93: I think you can also use "bigquery" format directly.

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -80,59 +107,65 @@ class AutoTask(
...
+        .option("path", mergePath)
+        .save()
 
-    val finalDataset = partitionedDF
-      .mode(task.write.toSaveMode)
-      .format(format.getOrElse(settings.comet.writeFormat))
-      .option("path", targetPath.toString)
+      val finalDataset = partitionedDF
+        .mode(task.write.toSaveMode)
+        .format(format.getOrElse(settings.comet.writeFormat))
```
AmineSagaama, line 128: We should use "temporaryGcsBucket" to save into BigQuery, unless we specify it in spark.conf.set(...)

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
hayssams: task.properties hold the properties we define like location, timestamp or days ttl.
The prefix spark will force AutoTask to add the properties to the session. Those properties are defined in the link https://github.com/GoogleCloudDataproc/spark-bigquery-connector but without the spark prefix.
This is where we define also the temporaryGcsBucket

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
hayssams: It did not work and a workaround on the Web was to directly reference the class.

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
hayssams: In this case we are writing to GCS not bigquery, wince target path is defined (line 112).
We return the data frame on line 169 and the effective write is done in INgestionWorkflow (line 405)

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -66,10 +67,36 @@ class AutoTask(
           .asInstanceOf[UdfRegistration]
       udfInstance.register(session)
     }
+    task.properties.foreach { properties =>
+      properties.map {
+        case (k, v) if k.startsWith("spark.") =>
+          session.conf.set(k.substring("spark.".length), v)
```
AmineSagaama, line 73: It's a good idea to use spark properties like this, it gives us more flexibility, however, we let the user override any spark properties in the yaml autojob file.

hayssams: You're right at least for this job

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -66,10 +67,36 @@ class AutoTask(
...
-        val df = session.read.parquet(fullPath)
+        val sepIndex = value.indexOf(":")
+        val (format, path) =
+          if (sepIndex > 0)
+            (IndexSink.fromString(value.substring(0, sepIndex)), value.substring(sepIndex + 1))
+          else // parquet is the default
+            (IndexSink.FS, value)
+
+        val df = format match {
+          case FS =>
```
AmineSagaama, line 87: Before adding bigquery feature, we use only parquet as source, so I think maybe in the pattern match we should use something like this : 
```
case BQ =>
            session.read
              .format("com.google.cloud.spark.bigquery")
              .option("table", path)
              .load()
              .cache()
case _ =>
val fullPath =
              if (path.startsWith("/")) path else s"${settings.comet.datasets}/$path"
            session.read.parquet(fullPath)
```

hayssams: Two lines above these ones  (line 84) , I force IndexSink.FS as being the default.

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
```diff
@@ -80,59 +107,65 @@ class AutoTask(
...
+      val partitionedDF =
+        partitionedDatasetWriter(
+          if (coalesce) dataframe.coalesce(1) else dataframe,
+          task.getPartitions()
+        )
+      partitionedDF
+        .mode(SaveMode.Overwrite)
+        .format(settings.comet.writeFormat)
+        .option("path", mergePath)
+        .save()
```
mhdkassir, line 124: this comment is not directly related to the PR , but it seems we are saving the the DF twice , I think we can remove saving to merge path which is not used :
```
      partitionedDF
      .option("path", mergePath)
        .mode(SaveMode.Overwrite)
      .save()
        .format(settings.comet.writeFormat)
        .option("path", mergePath)
        .save()
```
as well as this line :
`val _ = storageHandler.delete(new Path(mergePath))`
Only the `finalDataset` save is needed

src/main/scala/com/ebiznext/comet/workflow/IngestionWorkflow.scala
```diff
@@ -393,15 +392,18 @@ class IngestionWorkflow(
...
       )
       action.run() match {
-        case Success(_) =>
+        case Success(SparkJobResult(session, maybeDataFrame)) =>
           task.getIndexSink() match {
             case Some(IndexSink.ES) if settings.comet.elasticsearch.active =>
               index(job, task)
             case Some(IndexSink.BQ) =>
               val (createDisposition, writeDisposition) = Utils.getDBDisposition(task.write)
+              val sourceFile = maybeDataFrame
```
mhdkassir, line 401: perhaps rename this  val as `source`, having "file" in the name of the val can be misinterpreted

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
mhdkassir: the comment needs some typo correction :)

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
hayssams: Well spotted

src/main/scala/com/ebiznext/comet/workflow/IngestionWorkflow.scala
hayssams: Done

src/main/scala/com/ebiznext/comet/job/transform/AutoTask.scala
AmineSagaama: For the spark properties, maybe we should retrieve them from **dataproc_spark_properties** (in the task Dag) instead of reference them from the AutoJob yaml file.

hayssams: Good point. They will be automatically available from there too since qe are using a shared session for the whole job.

PR #263, opened 2020-06-15
BigQuery Connector does not support Dynamic Overwrite of Partitions

AmineSagaama:
## Summary

BigQuery connector should support dynamic overwrite of partitions : 
The newly inserted partitions would overwrite only partitions being inserted, and other partitions that were not part of that group would also stick around untouched.
But actually BigQuery connector does not support that right now.
The purpose of This PR is to fix this by using an APPEND mode in case of partitioning Table,
so old partitions remain intact.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryLoadJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryLoadJob.scala
@@ -97,9 +97,12 @@ class BigQueryLoadJob(
     }
 
     val writeDisposition = JobInfo.WriteDisposition.valueOf(cliConfig.writeDisposition)
-    val finalWriteDisposition = writeDisposition match {
-      case JobInfo.WriteDisposition.WRITE_TRUNCATE =>
-        logger.info(s"Deleting table $tableId")
+    val finalWriteDisposition = (writeDisposition, cliConfig.outputPartition) match {
+      case (JobInfo.WriteDisposition.WRITE_TRUNCATE, Some(partition)) =>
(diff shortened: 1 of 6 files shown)


src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryLoadJob.scala
```diff
@@ -170,16 +150,44 @@ class BigQueryLoadJob(
...
         s"BigQuery Saving to  ${table.getTableId} containing ${stdTableDefinition.getNumRows} rows"
       )
-      sourceDF.write
-        .mode(SaveMode.Append)
-        .format("com.google.cloud.spark.bigquery")
-        .option("table", bqTable)
-        .save()
+
+      (cliConfig.writeDisposition, cliConfig.outputPartition) match {
+        case ("WRITE_TRUNCATE", Some(partition)) =>
```
hayssams, line 162: Does not handle int partitions

src/main/scala/com/ebiznext/comet/workflow/IngestionWorkflow.scala
```diff
@@ -399,7 +401,7 @@ class IngestionWorkflow(
             case Some(IndexSink.BQ) =>
               val (createDisposition, writeDisposition) = Utils.getDBDisposition(task.write)
               val source = maybeDataFrame
-                .map(Right(_))
+                .map(df => Right(setNullableStateOfColumn(df, nullable = true)))
```
hayssams, line 404: Because the column is required in the Dataframe. We need to make it optional or else BQ would fail.

PR #265, opened 2020-06-16
WIP: Feature/use replace file

mhdkassir:
## Summary
Use hdfs `FileUtil.replaceFile` to move files instead of copying then deleting which can be expensive with big files.

**PR Type: Feature**

**Status: WIP**

**Breaking change? No**

### How has this been tested?
Existing Unit Tests

### Other changes
Add an evironements variable for tests to fix a a test that needs a project_id to complete successfully.

### Remaining Todos
* [ ] - Test an end to end ingestion on GCP

## Contributor checklist:

### DIFF
--- a/src/main/scala/com/ebiznext/comet/schema/handlers/StorageHandler.scala
+++ b/src/main/scala/com/ebiznext/comet/schema/handlers/StorageHandler.scala
@@ -20,7 +20,7 @@
 
 package com.ebiznext.comet.schema.handlers
 
-import java.io.ByteArrayOutputStream
+import java.io.{ByteArrayOutputStream, File}
 import java.time.{Instant, LocalDateTime, ZoneId}
 
 import com.ebiznext.comet.config.Settings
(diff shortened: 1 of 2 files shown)


src/main/scala/com/ebiznext/comet/schema/handlers/StorageHandler.scala
```diff
@@ -206,8 +206,20 @@ class HdfsStorageHandler(fileSystem: Option[String])(implicit
     * @return
     */
   override def move(path: Path, dest: Path): Boolean = {
-
-    FileUtil.copy(fs, path, fs, dest, true, true, conf)
+    val fileFromPath: Path => File = p =>
```
hayssams, line 209: This won't work on HDFS or GCS.
Why don't we simply use http://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FileSystem.html#rename-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Path-
We also should delete the dest file first if it already exist to keep the function semantic.

mhdkassir: done, I'll keep the PR in WIP until I do some tests on GCP

PR #281, opened 2020-06-26
Feature : Comet scripted fields

elarib:
## Summary
This PR add the possibility to add scripted fields in Comet.
Scripted fields compute data on the fly from the data being ingested.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### Solution
Now we can add the field scripted which is the sql part that create the field.  

Example :
 - id: long
 - name: String

In the domain we can have: 

      - name: "id"
        type: "long"
      - name: "name"
        type: "string"
      - name: nameUpperCase
        type: "string"
        scripted: "upper(name)"

Now in the output, we're going to have also a new column 'nameUpperCase' which contain the name upperCase
 
There was a problem when writing scripted: "input_file_name()", because when we apply this script we sometimes, already lost the inputFileName, so to fix that, we always now add a new column (comet_input_file_name) that we can use instead of input_file_name().

Please check the UT for more info.

### How has this been tested?
Just Unit tests for now

### Remaining Todos
* [x] - UT updated accordingly
* [ ] - Run it in a cluster

## Contributor checklist:
- [x] My change requires a change to the documentation.

### DIFF
--- a/src/test/resources/sample/simple-json-locations/locations.yml
+++ b/src/test/resources/sample/simple-json-locations/locations.yml
@@ -32,14 +32,14 @@ schemas:
         privacy: "NONE"
       - name: "name_upper_case"
         type: "string"
-        scripted: "upper(name)"
+        script: "upper(name)"
         array: false
         required: true
         privacy: "NONE"
         metricType: "NONE"
       - name: "source_file_name"
         type: "string"
-        scripted: regexp_extract(comet_input_file_name, '.+\/(.+)$' , 1)
+        script: regexp_extract(comet_input_file_name, '.+\/(.+)$' , 1)
         array: false
         required: true
         privacy: "NONE"
--- a/src/test/resources/sample/simple-json-locations/locations.yml
+++ b/src/test/resources/sample/simple-json-locations/locations.yml
@@ -30,3 +30,17 @@ schemas:
         type: "string"
         required: false
         privacy: "NONE"
+      - name: "name_upper_case"
+        type: "string"
+        scripted: "upper(name)"
+        array: false
+        required: true
+        privacy: "NONE"
+        metricType: "NONE"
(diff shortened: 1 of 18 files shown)


src/test/scala/com/ebiznext/comet/database/extractor/ExtractScriptGenSpec.scala
```diff
@@ -7,28 +7,28 @@ import org.scalatest.matchers.should.Matchers
...
-        getClass.getResource("/sample/database/EXTRACT_TABLE.sql.mustache").getPath
-      ),
-      templateParams
-    )
-
-    templatePayload shouldBe File(
-      getClass.getResource("/sample/database/expected_script_payload.txt").getPath
-    ).lines.mkString("\n")
-  }
+//
```
hayssams, line 10: Why did you comment this UT

hayssams:
@elarib Do you think you'll be able submit this one again soon.
We are all very excited to start using it.

elarib:
> @elarib Do you think you'll be able submit this one again soon.
> We are all very excited to start using it.

Yes, i just did. Please check :)

src/test/resources/sample/simple-json-locations/locations.yml
```diff
@@ -30,3 +30,17 @@ schemas:
...
+      - name: "name_upper_case"
+        type: "string"
+        scripted: "upper(name)"
+        array: false
+        required: true
+        privacy: "NONE"
+        metricType: "NONE"
+      - name: "source_file_name"
+        type: "string"
+        scripted: regexp_extract(comet_input_file_name, '.+\/(.+)$' , 1)
```
hayssams, line 42: It would be gr8 if we could rename scripted as script

PR #291, opened 2020-07-10
Giving the possibility to provide a mapping table name / delta column if

mmenestret:
Giving the possibility to provide a mapping table name / delta column if a single default col is not enough.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
+++ b/src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
@@ -2,6 +2,7 @@ package com.ebiznext.comet.database.extractor
 
 import better.files.File
 import com.ebiznext.comet.config.{DatasetArea, Settings}
+import com.ebiznext.comet.database.extractor.config.{Settings => ExtractorSettings}
 import com.ebiznext.comet.schema.handlers.SchemaHandler
 import com.ebiznext.comet.schema.model.Domain
 import com.typesafe.config.ConfigFactory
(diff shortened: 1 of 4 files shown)


src/main/scala/com/ebiznext/comet/database/extractor/ExtractScriptGen.scala
```diff
@@ -30,15 +31,18 @@ object ScriptGen extends StrictLogging {
...
+    * @param deltaColumns Mapping table name -> delta column, has precedence over `defaultDeltaColumn`
     * @return The list of produced files
     */
   def generate(
     domain: Domain,
     scriptTemplateFile: File,
     scriptsOutputPath: File,
-    deltaColumn: Option[String]
+    defaultDeltaColumn: Option[String],
+    deltaColumns: Map[String, String]
```
AmineSagaama, line 43: Is it enough to have only one delta column for each Table ? Maybe some Tables need a list of columns to apply a diff ?

hayssams: I do not yet understand this PR 
Being in front of the beach does not help
But usually it’s better to allow the user to use multiple columns

cchepelov:
Meta-review: why are you staring at PRs while at the beach? 😜

hayssams:
image

Because drinking wine and eating oysters does not make you clever

AmineSagaama:
@mmenestret 
Sounds good to me, to manage multiple columns in another PR.

PR #296, opened 2020-07-17
Consider cells with whitespaces only as blank cells

mmenestret:
## Summary
When parsing Excel cells, consider cells with only whitespaces as blank to avoid human mistakes.

### DIFF
--- a/src/main/scala/com/ebiznext/comet/schema/generator/XlsReader.scala
+++ b/src/main/scala/com/ebiznext/comet/schema/generator/XlsReader.scala
@@ -5,10 +5,9 @@ import java.util.regex.Pattern
 
 import com.ebiznext.comet.config.Settings
 import com.ebiznext.comet.schema.model._
-import org.apache.poi.ss.usermodel.{Cell, DataFormatter, Row, Workbook, WorkbookFactory}
+import org.apache.poi.ss.usermodel._
 
 import scala.collection.JavaConverters._
-import scala.util.{Success, Try}
 
 /**
   * Reads the spreadsheet found at the specified {@param path} and builds the corresponding Domain object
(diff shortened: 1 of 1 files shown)


src/main/scala/com/ebiznext/comet/schema/generator/XlsReader.scala
```diff
@@ -18,21 +17,27 @@ class XlsReader(path: String) {
...
     workbook.getSheet("domain").asScala.drop(1).headOption.flatMap { row =>
       val nameOpt = Option(row.getCell(0, Row.MissingCellPolicy.RETURN_BLANK_AS_NULL))
-        .map(formatter.formatCellValue)
+        .flatMap(formatter.formatCellValue)
       val directoryOpt = Option(row.getCell(1, Row.MissingCellPolicy.RETURN_BLANK_AS_NULL))
-        .map(formatter.formatCellValue)
+        .flatMap(formatter.formatCellValue)
       val ack = Option(row.getCell(2, Row.MissingCellPolicy.CREATE_NULL_AS_BLANK))
-        .map(formatter.formatCellValue)
+        .flatMap(formatter.formatCellValue)
```
mhdkassir, line 38: Please make sure that `ack `is still set to "" in the generated yaml when the cell is empty.

mmenestret: I'll double check that behavior

mmenestret: Don't you think that:

```def getAck(): String = ack.map(ack => if (ack.nonEmpty) "." + ack else ack).getOrElse(".ack")```

Is enough ?

hayssams:
Why is this PR still waiting to be merged ?

PR #297, opened 2020-07-20
Timestamped csv

hayssams:
## Summary
Sometimes we want to ingest a file and save it in a single CSV file. This should never happen except at the preencrypt phase before the file is sent for ingestion.
We achieve this by setting the timestamped-csv property to true in the reference.conf or by using the COMET_TIMESTAMPED_CSV env var.
This feature requires partitioning and grouped ingestion to be disabled "grouped = ${?COMET_GROUPED}" otherwise it is ignored.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? Yes**

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -85,6 +85,9 @@ trait IngestionJob extends SparkJob {
       .map(_ => WriteMode.OVERWRITE)
       .getOrElse(metadata.getWriteMode())
 
+  private def timestamedCsv(): Boolean =
+    settings.comet.timestampedCsv && !settings.comet.grouped && metadata.partition.isEmpty
+
   /**
     * Merge new and existing dataset if required
     * Save using overwrite / Append mode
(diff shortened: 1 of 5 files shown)


src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -84,6 +85,9 @@ trait IngestionJob extends SparkJob {
       .map(_ => WriteMode.OVERWRITE)
       .getOrElse(metadata.getWriteMode())
 
+  private def timestamedCsv(): Boolean =
```
mhdkassir, line 88: typo correction : timestamped

hayssams: done

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -382,10 +400,18 @@ trait IngestionJob extends SparkJob {
...
-      val finalTargetDatasetWriter = targetDatasetWriter
-        .mode(saveMode)
-        .format(settings.comet.writeFormat)
-        .option("path", targetPath.toString)
+      val finalTargetDatasetWriter =
+        if (timestamedCsv())
+          targetDatasetWriter
+            .mode(saveMode)
+            .format("csv")
+            .option("path", targetPath.toString)
```
mhdkassir, line 408: add two options for header and delimiter to be have an output CSV in compliance with the schema specifications
```
            .option("header", metadata.withHeader.getOrElse(false))
            .option("delimiter", metadata.separator.getOrElse("µ"))
```

hayssams: Done

PR #304, opened 2020-07-24
Support BigQuery on merge.

hayssams:
## Summary
Support BigQuery on merge.
When merging during ingestion, if the file is written to BQ, apply the merge strategy on the BigQuery table directly.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/ingest/DsvIngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/DsvIngestionJob.scala
@@ -150,7 +150,7 @@ class DsvIngestionJob(
             df.select(
               cols.head,
               cols.tail
-                .take(attributesWithoutscript.length - 1) :+ Settings.cometInputFileNameColumn: _*
+                .take(attributesWithoutscript.length - 1): _*
             ).toDF(attributesWithoutscript.map(_.name): _*)
           }
       }
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -137,17 +137,34 @@ trait IngestionJob extends SparkJob {
 
                                       } else acceptedDF).drop(Settings.cometInputFileNameColumn)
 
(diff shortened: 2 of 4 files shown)


src/main/scala/com/ebiznext/comet/job/ingest/DsvIngestionJob.scala
```diff
@@ -150,7 +150,7 @@ class DsvIngestionJob(
             df.select(
               cols.head,
               cols.tail
-                .take(attributesWithoutscript.length - 1) :+ Settings.cometInputFileNameColumn: _*
```
hayssams, line 153: This si a bug. It generates an exception and occur only for CSV files without header.
This should not be done here since the file is not yet loaded at this step.

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -137,17 +137,34 @@ trait IngestionJob extends SparkJob {
 
                                       } else acceptedDF).drop(Settings.cometInputFileNameColumn)
 
-    val mergedDF = schema.merge.map { mergeOptions =>
-        if (storageHandler.exists(new Path(acceptedPath, "_SUCCESS"))) {
+    val mergedDF = schema.merge
+      .map { mergeOptions =>
+        if (metadata.getIndexSink().getOrElse(IndexSink.None) == IndexSink.BQ) {
+          // When merging to BigQuery, load existing DF from BigQuery
+          val table = BigQueryLoadJob.getTable(session, domain.name, schema.name)
```
AmineSagaama, line 144: Could you explain why we have to call BigQuery Client API to retrieve the metadata of the Table ? 
Because, to merge the Dataframe and save it to BigQuery Table, you can directly load the Table as a Dataframe,
apply the merge, and save it as a Table.

hayssams: I need to check that the table exists.
When the table doe snot exist, it means that this is a first time ingestion and no merge is needed.

PR #308, opened 2020-07-24
Spark 3 Support

hayssams:
## Summary
Support Spark 3 with scala 2.12 et Spark with Scala 2.11 and 2.12.
By default set will build with scala 2.12 and Spark3 
To force a specific version simply pass the Spark version in an env var as described below :

COMET_SPARK_VERSION=2.4.6 sbt ++2.11.10 assembly
COMET_SPARK_VERSION=2.4.6 sbt ++2.12.12 assembly
COMET_SPARK_VERSION=3.0.0 sbt ++2.12.12 assembly

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? Yes**
The resulting assembly is now named "com-spark2" or "comet-spark3". It was originally named "comet"


project/Common.scala
```diff
@@ -104,6 +104,7 @@ object Versions {
   val spark211 = "2.1.0"
   val spark211_240 = "2.4.6"
   val spark212 = "2.4.6"
+  val spark312 = "3.0.0"
```
cchepelov, line 107: why is this named "spark312" ? Original naming convention did suggest "spark212 = spark for scala 2.12" but this...

How about turning around the whole thing in the admittedly more verbose but possibly clearer in intent (yes, supporting version matrices is always awkward but there isn't really a way around that, is there?)

spark211 -> spark_2d1_forScala_2d11 
jackson212 -> jackson_forSpark_2d1_forScala_2d11 
jackson312 -> jackson_forSpark_3   // we don't (yet and hopefully never) support different versions for different scalas on spark 3

sbt-assembly-all.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+#COMET_SPARK_VERSION=2.1.0 sbt ++2.11.10 assembly
+COMET_SPARK_VERSION=2.4.6 sbt ++2.11.10 assembly
```
cchepelov, line 4:
```suggestion
COMET_SPARK_VERSION=2.4.6 sbt ++2.11.12 assembly
```

Scala 2.11.10 is really ancient and there is a CVE open against it. Scala 2.11.12 should be the one flintstone we need.

build.sbt
```diff
@@ -19,27 +23,48 @@ organization := "com.ebiznext"
...
+scalaVersion := scala212
 
 organizationHomepage := Some(url("http://www.ebiznext.com"))
 
+val sparkVersionPattern: Regex = "(\\d+).(\\d+).(\\d+)".r
+val sparkPatternMatch = sparkVersionPattern
+  .findFirstMatchIn(sparkVersion)
+  .getOrElse(throw new Exception(s"Invalid Spark Version $sparkVersion"))
+val sparkMajor = sparkPatternMatch.group(1)
+val sparkMinor = sparkPatternMatch.group(2)
```
cchepelov, line 35:
```suggestion
val (sparkMajor, sparkMinor) = sparkVersion match {
  case sparkVersionPattern(maj, min) => 
    (maj, min)
  case _ =>
      throw new IllegalArgumentException(s"Unexpected Spark Version $sparkVersion")
}
```

hayssams:
Will correct it as suggested ASAP

build.sbt
```diff
@@ -19,27 +23,48 @@ organization := "com.ebiznext"
 
 organizationName := "Ebiznext"
 
-scalaVersion := scala211
+scalaVersion := scala212
 
 organizationHomepage := Some(url("http://www.ebiznext.com"))
 
+val sparkVersionPattern: Regex = "(\\d+).(\\d+).(\\d+)".r
```
hayssams, line 30: Unfortunately Pattern matching in val statements is not supported

cchepelov:
Damn, right

build.sbt
cchepelov: ( https://github.com/sbt/sbt/issues/2290 FYI — a regression from sbt 0.13.6 (!) which has been festering for 5 years now... not the first time I got bit by that )

hayssams:
@cchepelov Th travis update looks perfect

cchepelov:
Thanks! apologies for the noise, I didn't expect to stomp onto your branch for that long.. you'll probably want to do a "rebase -i" before merging ;)

hayssams:
Actually Spark 2.1 support was partly dropped and wz were recompiling comet at the customer site.
Since you made the build more inclusive, I updated the code to keep the support for Spark 2.1 working

cchepelov:
(note that we *could* silence those warnings using Silencer from https://github.com/ghik/silencer)

Seems (from the allowed failures) that Spark 2.4.6 still uses ASM6 on scala 2.11, and only switched up on scala 2.12; for that reason, I don't think we can support JDK>8 on scala 2.11 (JRE11 to run a JDK8-built assembly might be fine, but caveat emptor)

hayssams:
Since the only customer using Spark 2.1 is moving toward Spark 2.4 in September, I too think that it’s fine.

Issue #289, opened 2020-07-06
[BUG] - Encoding ignored for Position files


mhdkassir:
## Description
A Position file with an encoding different than UTF-8 is not parsed correctly.

## Expected behavior
Given a Position file with an encoding different than UTF-8, say iso-8859-1 for example. We can set the "encoding" parameter in the Schema metadata, and Spark should be able to read the data correctly especially when it comes to accents and special characters. 
 
## Current Behavior
Even when the encoding parameter is set with the right Charset, the resulting DataFrame when loading a Position file with an iso-8859-1 encoding contains a wrong representation of special characters (�). 
This is due to the implementation of `PositionIngestionJob.loadDataSet()` where we have the following:
`session.read.option("encoding", metadata.getEncoding()).text(...)`
The `text() `method does not support the "encoding" option (only available for `csv() `). It only reads UTF-8 charset.
A spark issue exists for this matter and it was resolved with a "Won't fix" resolution : https://issues.apache.org/jira/browse/SPARK-1849

## Possible implementation
_TODO_

## Complexity estimation
T-Shirt Size: XS, S, M, L. If you would recommend XL or more, we advise you to discuss with the community how to split this issue.
Don't forget to tag this issue as a 'good first issue' if you think it is appropriate!

PR #309, opened 2020-07-26
Bugfix/encoding for position files

mhdkassir:
## Summary
Adds a new method to read non-csv text files with an encoding different than UTF-8

**Related Issue: #289**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
`PositionIngestinJob.loadDataSet() `checks for schema encoding before reading the data. When the encoding is different than UTF-8, data is loaded using the method `PositionIngestionUtil.loadDfWithEncoding` which can take in consideration the specified encoding.

### How has this been tested?
Adding a new unit test

### Other changes
minor correction in Domain validation error message

## Contributor checklist:


src/main/scala/com/ebiznext/comet/job/ingest/PositionIngestionJob.scala
```diff
@@ -57,8 +59,14 @@ class PositionIngestionJob(
     */
   override def loadDataSet(): Try[DataFrame] = {
     try {
-      val df =
-        session.read.option("encoding", metadata.getEncoding()).text(path.map(_.toString): _*)
```
hayssams, line 61: I don't understand why setting the option to "ISO-8859-1" in the encoding property won't load the file in the specified encoding ? Is it a Spark bug ?

hayssams: Looking ag this PR https://github.com/apache/spark/pull/21287
I feel like there is an easier way too support custom encoding in non csv text file

hayssams: unfortunately this PR was marked as Won't Fix

mhdkassir: > unfortunately this PR was marked as Won't Fix

exactly, I found another PR on spark about this subject that I indicated in the bug description (https://issues.apache.org/jira/browse/SPARK-1849) and it was also marked as _Won't Fix_

PR #314, opened 2020-07-28
Clustering Support for BigQuery Tables

hayssams:
## Summary
Add an extra option to the metadata properties to allow user to specify clustering option.
We probably need to unify partition & clustering options with Hive./
The new property is named clustering and is specified as a comma separated list of columns 

"index" is renamed "sink" and sink properties are located in the sink bloc.
Sink properties are sink type dependant.

Below are JDBC / BQ and ES sinktypes.
```
  sink:
    type: ES
    timestamp: "{beginTs|yyyy.MM.dd}"
    days: 7

  sink:
    type: BQ
    location: EU
    timestamp: _PARTITIONTIME
    days: 7
    require-partition-filter: true


  sink:
    type: JDBC
    connection: postgres
    partitions: 1
    batchsize: 1000
```

We now use the exact same class to load sink config from reference.conf and from the Metadata section in the YML files.
Both serve the same purpose except that they are read by two different libs "kxbmap Configs" vs Jackson YAML 
Jackson allows us to use friendly names from the type field while kxbmap require the case class name in the type field.
Thus to describe a sink in the conf file we need to use the names NoneSink, JdbcSink, BigQuerySink, EsSink and in the YML files we need to use the friendly names None, BQ, JDBC, ES

**PR Type: Feature**

**Status: Ready**

**Breaking change? Yes**


src/main/scala/com/ebiznext/comet/schema/model/Metadata.scala
```diff
@@ -177,73 +169,3 @@ object Metadata {
       None
     )
 }
-
-class MetadataDeserializer extends JsonDeserializer[Metadata] {
```
mhdkassir, line 181: Why deleting this class, isn't there any impact ?

src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
```diff
@@ -0,0 +1,120 @@
...
+package com.ebiznext.comet.schema.model
+
+import com.fasterxml.jackson.annotation.{JsonSubTypes, JsonTypeInfo, JsonTypeName}
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.annotation.{JsonDeserialize, JsonSerialize}
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer
+import com.fasterxml.jackson.databind.{DeserializationContext, JsonDeserializer}
+
+/**
+  * Recognized file type format. This will select  the correct parser
```
mhdkassir, line 30: scaladoc does not correspond SinkType

mhdkassir:
Given the breaking changes in generating yml files, I think it would be safer to release the version 0.11.0 from master branche before merging this PR, and then merge on the version 0.12.0-SN. This would allow a smooth switching to the new structure of `Metadata` and `Sink`

hayssams:
@mhdkassir current master published for Spark 2 -> Scala 2.11 / Scala 2.12

PR #319, opened 2020-08-04
Ignore lines using regex / UDF

hayssams:
## Summary
Add the ignore key to the metadata section; This will allow us to remove lines matching a regex or that match an UDF condition.

The syntax is as follow :
```
metadata:
  ignore: "IGNORE.*"
```

Where "IGNORE.*" is a regex

or
```
metadata:
  ignore: "udf:ignoreUdf"
```

where ignoreUdf is a registered UDF through the "udfs" key in the reference.conf file with the following signature: (r: Row) => Boolean.

See an example in com.ebiznext.comet.udf.TestUdf.

Also Allow the user to define UDFs at the reference.conf level through the "udfs" key as a comma separated list of UDFRegistration classes. See com.ebiznext.comet.udf.TestUdf

This key may also be defined through the COMET_UDFS environment variable.

UDFs defined here are available in all ingestion / Auto Jobs throughout the project.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -62,6 +62,18 @@ trait IngestionJob extends SparkJob {
     */
   def ingest(dataset: DataFrame): (RDD[_], RDD[_])
 
+  def applyIgnore(dfIn: DataFrame): Dataset[Row] = {
+    import org.apache.spark.sql.functions._
+    import session.implicits._
+    metadata.ignore.map { ignore =>
+      if (ignore.startsWith("udf:")) {
+        dfIn.filter(callUDF(ignore.substring("udf:".length), struct(dfIn.columns.map(dfIn(_)): _*)))
+      } else {
+        dfIn.filter(!($"value" rlike ignore))
+      }
+    } getOrElse dfIn
+  }
+
   def saveRejected(rejectedRDD: RDD[String]): Try[Path] = {
     logger.whenDebugEnabled {
       logger.debug(s"rejectedRDD SIZE ${rejectedRDD.count()}")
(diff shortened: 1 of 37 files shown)


hayssams:
This PR requires the PR UDF to be merged first

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -62,14 +62,26 @@ trait IngestionJob extends SparkJob {
     */
   def ingest(dataset: DataFrame): (RDD[_], RDD[_])
 
+  def applyIgnore(dfIn: DataFrame): Dataset[Row] = {
+    import org.apache.spark.sql.functions._
+    import session.implicits._
+    metadata.ignore.map { ignore =>
+      if (ignore.startsWith("udf:")) {
+        dfIn.filter(callUDF(ignore.substring("udf:".length), struct(dfIn.columns.map(dfIn(_)): _*)))
```
mhdkassir, line 70: should apply a filterNot

src/main/scala/com/ebiznext/comet/schema/model/Metadata.scala
```diff
@@ -135,9 +141,32 @@ case class Metadata(
       escape = merge(this.escape, child.escape),
       write = merge(this.write, child.write),
       partition = merge(this.partition, child.partition),
-      sink = merge(this.sink, child.sink)
+      sink = merge(this.sink, child.sink),
+      ignore = merge(this.ignore, child.ignore)
     )
   }
+
+  def checkValidity(
```
mhdkassir, line 149: this method seems to only be called in tests , it is not intended to be used to validate `Metadata` during ingestion ?

PR #326, opened 2020-08-12
BUG FIX - Using the ingested df schema when merging with an existing source

mmenestret:
## Summary
Using the ingested df schema when merging with an existing source

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

## Description

When retrieving the DF from BQ, it messes up with the nullable attribute of each field (all fields are considered nullable in the loaded DF).
When joined with the ingested DF, we end up with the schema of the DF loaded from BQ (being on the left part of the join I guess).
It results in a rejection by BQ when trying to write the table (provided schema does not match Table ... . Field XXX has changed mode from REQUIRED to NULLABLE).


mhdkassir:
> It seems there is a regression in some Unit Tests, `SchemaHandlerSpec` at least

@mmenestret  more precisely the failing test is _"ingest schema with merge"_

mmenestret:
I found what went wrong @mhdkassir .
When, during tests, we persist write the dataset as a parquet file, we use 2 cols for partitioning
```
|-- YEAR: string (nullable = true)
|-- MONTH: string (nullable = true)
```
which are defined that way in the schema
```
      - name: "YEAR"
        type: "string"
        array: false
        required: false
        privacy: "NONE"
      - name: "MONTH"
        type: "string"
        array: false
        required: false
        privacy: "NONE"
```.

When the dataframe writer write it down on the merge path, those 2 cols turns into int columns (??? no idea why), and then, when trying to apply the new ingested file schema (which has those cols as string) to it (the new createDataframe call I added), it fails since the existing df cols are now int.

Do you have any idea ?

**Comment by mmenestret** [comment]
Here is an illustration of what seems to be wrong:

```
scala> case class Test(name: String, day: String, month:String)
defined class Test

scala> case class Test(name: String, day: String)
defined class Test

scala> val l = List(Test("a", "1"), Test("b", "2"))
l: List[Test] = List(Test(a,1), Test(b,2))

scala> spark.createDataFrame(l)
res9: org.apache.spark.sql.DataFrame = [name: string, day: string]

scala> res9.printSchema
root
 |-- name: string (nullable = true)
 |-- day: string (nullable = true)

scala> res9.write.partitionBy("day").parquet("/tmp/test")

scala> val df2 = spark.read.parquet("/tmp/test")
df2: org.apache.spark.sql.DataFrame = [name: string, day: int]

scala> df2.printSchema
root
 |-- name: string (nullable = true)
 |-- day: integer (nullable = true)
```

**Comment by mhdkassir** [comment]
> Here is an illustration of what seems to be wrong:
> 
> ```
> scala> case class Test(name: String, day: String)
> defined class Test
> 
> scala> val l = List(Test("a", "1"), Test("b", "2"))
> l: List[Test] = List(Test(a,1), Test(b,2))
> 
> scala> spark.createDataFrame(l)
> res9: org.apache.spark.sql.DataFrame = [name: string, day: string]
> 
> scala> res9.printSchema
> root
>  |-- name: string (nullable = true)
>  |-- day: string (nullable = true)
> 
> scala> res9.write.partitionBy("day").parquet("/tmp/test")
> 
> scala> val df2 = spark.read.parquet("/tmp/test")
> df2: org.apache.spark.sql.DataFrame = [name: string, day: int]
> 
> scala> df2.printSchema
> root
>  |-- name: string (nullable = true)
>  |-- day: integer (nullable = true)
> ```

as a workaround I would suggest to change the type of year and month in the test file (DOMAIN.yml) to int. In reality I think most of the times the partition column type will be an integer and not a string, especially for values such as month, day, and year. What do you think @mmenestret

**Comment by mmenestret** [comment]
Well, there are 2 options:
- Either we enforce partition columns to be int columns, but there might still be some problems, we need them to be int _primitive_ type not semantic type, e.g. a type.yml like this would make it fail as well since big int cannot be cast to int
```
- name: "int"
    pattern: "[-|\\+|0-9][0-9]*"
    primitiveType: "long"
    sample: "1234"
    comment: "Int number"
```

- Or, when reading existing DF on disk, we provide the current DF schema :

```
else if (storageHandler.exists(new Path(acceptedPath, "_SUCCESS"))) {
          // Otherwise load from accepted area
          val existingDF =
            session.read.schema(acceptedDfWithscriptFields.schema).parquet(acceptedPath.toString)
          merge(acceptedDfWithscriptFields, existingDF, mergeOptions)
        }
``` (just tested, it does fix the UT)

**Comment by mhdkassir** [comment]
> Some docs about the topic :
> 
>     * > Notice that the data types of the partitioning columns are automatically inferred. Currently, numeric data types, date, timestamp and string type are supported. Sometimes users may not want to automatically infer the data types of the partitioning columns. For these use cases, the automatic type inference can be configured by spark.sql.sources.partitionColumnTypeInference.enabled, which is default to true. When type inference is disabled, string type will be used for the partitioning columns.
>       > ([Spark doc](https://spark.apache.org/docs/2.4.0/sql-data-sources-parquet.html#partition-discovery))
> 
>     * https://stackoverflow.com/questions/46657954/avoid-losing-data-type-for-the-partitioned-data-when-writing-from-spark
> 
> 
> Well, there are 2 options:
> 
>     1. Either we enforce partition columns to be int columns, but there might still be some problems, we need them to be int _primitive_ type not semantic type, e.g. a type.yml like this would make it fail as well since big int cannot be cast to int
> 
> 
> ```
> - name: "int"
>     pattern: "[-|\\+|0-9][0-9]*"
>     primitiveType: "long"
>     sample: "1234"
>     comment: "Int number"
> ```
> 
>     1. Or, when reading existing DF on disk, we provide the current DF schema :
> 
> 
> ```
> else if (storageHandler.exists(new Path(acceptedPath, "_SUCCESS"))) {
>           // Otherwise load from accepted area
>           val existingDF =
>             session.read.schema(acceptedDfWithscriptFields.schema).parquet(acceptedPath.toString)
>           merge(acceptedDfWithscriptFields, existingDF, mergeOptions)
>         }
> ```
> 
> (just tested, it does fix the UT)
> 
> What do you think ?
> 
> BTW, what I suggested with that PR was:
> 
> ```
> val orderedExisting = session.createDataFrame(
>       existingDF.select(partitionedInputDF.columns.map(col): _*).rdd,
>       partitionedInputDF.schema
>     )
> ```
> 
> Which is trying to coerce the existing DF, after being loaded, to a different schema, which fails (string partition cols cannot be changed to int).
> It does not fail if the schema is given at load time (which is what I suggested as option 2 here).

providing the schema when reading the existingDF is a good idea too, but what happens when the schema changes ? Merge is not possible in this case, and so far this was managed at line 300 in IngestionJob :
```
    if (existingDF.schema.fields.length != partitionedInputDF.schema.fields.length) {
      throw new RuntimeException(
        "Input Dataset and existing HDFS dataset do not have the same number of columns. Check for changes in the dataset schema ?"
      )
    }
```
but with this proposition the test done here becomes useless. We will have a failure at runtime I think.

mhdkassir (approved):
LGTM

PR #327, opened 2020-08-12
BUG FIX - Allow BQ merge when there is no specified partitionning col

mmenestret:
## Summary
Allow BQ merge when there is no specified partitionning col

**Related Issue: #IssueId**

**PR Type: Bug Fix**

**Status: Ready to review**


AmineSagaama (approved):
LGTM

PR #332, opened 2020-09-01
replace timestampedCsv mode by csvOutput

mhdkassir:
## Summary
csvOutput mode saves the result in a csv file with the same name of the input file

**PR Type: Feature**

**Status: Ready To Review**

**Breaking change? No**

### How has this been tested?
- a local test for watch step with timestampedCsv option
- existing unit tests

### Other changes
- update an error message in DsvIngestionJob
- bugfix for pre and post encrypt generated yml

## Contributor checklist:


src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -210,7 +214,7 @@ trait IngestionJob extends SparkJob {
         .head
       val finalCsvPath = new Path(
         acceptedPath,
-        s"${acceptedPath.getName()}-${now}.csv"
+        s"${acceptedPath.getName()}-${now}${path.headOption.map("-" + _.getName).getOrElse(".csv")}"
```
hayssams, line 217: Shouldn't we set the timestamp at the end of the filename ?

src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
```diff
@@ -21,19 +21,23 @@ object Xls2Yml extends LazyLogging {
...
-            attr.copy(`type` = "string", required = false, privacy = None)
-        }
-      val newMetaData: Option[Metadata] = s.metadata.map { m => m.copy(partition = None) }
-      s.copy(attributes = newAtt)
+              attr.copy(`type` = "string", required = false)
+            else
+              attr.copy(`type` = "string", required = false, privacy = None)
+          }
+      val newMetaData: Option[Metadata] = s.metadata.map { m =>
+        m.copy(partition = None).copy(sink = None)
```
hayssams, line 38: m.copy(partition = None, sink = None) would be simpler to read

src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
```diff
@@ -21,19 +21,23 @@ object Xls2Yml extends LazyLogging {
...
+              )
             )
-          )
-            attr.copy(`type` = "string", required = false)
-          else
-            attr.copy(`type` = "string", required = false, privacy = None)
-        }
-      val newMetaData: Option[Metadata] = s.metadata.map { m => m.copy(partition = None) }
-      s.copy(attributes = newAtt)
+              attr.copy(`type` = "string", required = false)
```
hayssams, line 33: I do not understand what is the purpose of this change

src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
```diff
@@ -63,7 +67,8 @@ object Xls2Yml extends LazyLogging {
         metadata.copy(
           format = Some(Format.DSV),
           separator = delimiter.orElse(schema.metadata.flatMap(_.separator)).orElse(Some("µ")),
-          withHeader = schema.metadata.flatMap(_.withHeader)
+          withHeader = schema.metadata.flatMap(_.withHeader),
+          encoding = None
```
hayssams, line 71: Is it required to set encoding to None

mhdkassir: The encoding in the post-encrypt yml should be forced to UTF-8, in case we used a different encoding in preEncryption phase. I used None because UTF-8 is the default value. But you're right it is a bit confusing. fixed.

src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
mhdkassir: fixed

src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
mhdkassir: the main change is to ignore scripted attributes in the pre-encrypt yml (line 26), the rest is just spacing code format

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
mhdkassir: No need for timestamp after all, this mode is renamed to csvOutput. The output file has the same name of the input file.

hayssams (approved):

PR #335, opened 2020-09-05
Spark no more required when running BQ Job

hayssams:
## Summary
When running an AutoJob, it is now possible to specify one of the two engine : SPARK or BQ.
When BQ is specified, no Dataproc cluster is required. The AutoJob is run against BQ using the BQ API.

udf if used  should refer to a cloud storage location where the udfs are stored.
Views are created but should be explicitly deleted in the postsql statements.
PreSQL & PostSQL statements are always executed even if the main SQL fails.
Only the results of the main SQL are stored to the output Table.

**PR Type:  Feature**

**Status: WIP**

**Breaking change? No**


src/main/scala/com/ebiznext/comet/job/metrics/MetricsJob.scala
```diff
@@ -244,14 +236,14 @@ class MetricsJob(
...
+              None
             }
 
           case None =>
-            Success(SparkJobResult(session))
+            Success(None)
         }
     }
-    combinedResult.find(_.isFailure).getOrElse(Success(SparkJobResult(session)))
+    combinedResult.find(_.isFailure).getOrElse(Success(None)).map(SparkJobResult(_))
```
mhdkassir, line 246: simplify using a forall

mhdkassir (approved):

Issue #339, opened 2020-09-13
Support for databricks


ashitabh:
## Description
Is Comet data pipeline supported on databricks ( Azure ) 

## Expected behavior
Spark submit should support job execution

## Current Behavior
I tries to create a job using Spark submit , but it failed with following error 

## Steps to reproduce
Login to azure databricks 
1. Go to 'Jobs'
2. Create new job with jar "comet-spark3_2.12-0.1.14-SNAPSHOT-assembly.jar" uploaded on dbfs
3. run job from portal
4. See error ( Error log attached)
[stderr (1).txt](https://github.com/ebiznext/comet-data-pipeline/files/5213541/stderr.1.txt)

 OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=512m; support was removed in 8.0
Warning: Ignoring non-Spark config property: eventLog.rolloverIntervalSeconds
Exception in thread "main" java.lang.Exception: Failed to load config: ConfigError(Exceptional(java.lang.NoSuchMethodError: com.typesafe.config.Config.hasPathOrNull(Ljava/lang/String;)Z,List(extract)),Vector(Exceptional(java.lang.NoSuchMethodError: com.typesafe.config.Config.hasPathOrNull(Ljava/lang/String;)Z,List(extract, metrics)), Exceptional(java.lang.NoSuchMethodError:

hayssams:
@ashitabh  Thanks for reporting the issue. We just spotted how it happens.

@cchepelov @mhdkassir We have a dependency from a kbxmap.configs to a more recent version of the typesafe config.
Databricks 7.2 still uses type.config.1.2 where kbxmap.configs brings in the 1.3 version as shown in the dependency tree below
```
+-com.github.kxbmap:configs_2.12:0.4.4 [S]
| +-com.typesafe:config:1.3.1
````

We probably need to force the dependency to the older version or else rewrite the package name using some shading rules
What do you suggest ?

cchepelov:
Short term I'd say shading would be the way to go; it'd be a shame if one 
platform was able to force us to use obsolete libraries.



Le 13 septembre 2020 09:42:26 Hayssam Saleh <notifications@github.com> a 
écrit :

hayssams:
Just created a PR.
We probably need to do some testing on Databricks from now on.
I'll check how we can gain access to a Databricks testing platform

ashitabh:
A short term fix I found on github is to upgrade typesafe on databricks using init script 
[...]

PR #340, opened 2020-09-13
Shade typesafe config

hayssams:
## Summary
kxbmap bring in a recent version of typesafeconfig (1.3.1) while Databricks still require an older version (1.2.1).
Closing issue #339 

**Related Issue: #IssueId**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
By shading the typesafe.config lib, we solve the conflict.

### How has this been tested?
Not tested on Databricks.


project/Common.scala
```diff
@@ -106,7 +106,6 @@ object Versions {
   val spark3d0 = "3.0.0"
   val scalatest = "3.2.2"
   val springBoot = "2.0.6.RELEASE"
-  val typesafeConfig = "1.2.1"
```
mhdkassir, line 109: instead of removing this version value, wouldn't it be better to use it in an explicit dependency definition of _typesafeconfig_ ? eventually with a newer version, 1.3.1 for example.

cchepelov:
Agreed with @mhdkassir

project/Common.scala
hayssams: Done

mhdkassir (approved):
Lgtm

PR #346, opened 2020-09-17
Feature/multiple rls

hayssams:
## Summary
Allow the user to specify multiple RLS for each BQ Table

**PR Type: Feature **

**Status: Ready to review**

**Breaking change? No**


src/main/scala/com/ebiznext/comet/schema/model/RowLevelSecurity.scala
```diff
@@ -26,7 +26,7 @@ case class RowLevelSecurity(
 object RowLevelSecurity {
 
   def parse(input: String): RowLevelSecurity = {
-    val components = input.split(',')
+    val components = input.split('/')
```
mhdkassir, line 29: is there any reason to replace comma by slash as a separator ?

hayssams: Yes you're right 
I should have documented this one.
Since we have a liste of RLS in the CLI, they have to be separated by a comma (this is the way scopt works). Since a RLS is made of three parts separated by a comma, we have to switch to a different char. 
That's what has been done here.

mhdkassir (approved):
LGTM

PR #350, opened 2020-09-18
Bugfix : several bug fixes

mhdkassir:
## Summary
- Remove a hard coded return value from `SimpleLauncher.ingest`
- Checking if input directory exists should be done in import step instead of watch
- Partition block should not be generated in yml when neither `sampling` nor `attributes` are set in xls file

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
cf. _Summary_

### How has this been tested?
- Existing Unit Tests
- Local testing

## Contributor checklist:


hayssams (approved):

hayssams:
LGTM

PR #353, opened 2020-09-29
Predicate Pushdown When Reading BigQuery Table

AmineSagaama:
## Summary
Spark BigQuery Connector is not able to pushdown predicate, when we're reading BigQuery Tables.
We Have to specify a **filter** option in the following manner:
```
val df = spark.read.format("com.google.cloud.spark.bigquery")
  .option("filter", "_PARTITIONDATE > '2019-01-01'")
  ...
  .load(TABLE)
```
So, from now on, we have two modes:
1/ **Standard Mode** : This is the current mode, we don't have to specify a filter here, but, we will load the entire table, so we risk to inflate the BigQuery bill.
2/ **Filter Mode**: Here, we could specify a set of filters for each input table.
If the input tables are partitioned, we should only select a specific partition, therefore, we optimize our BigQuery costs.

Here is an exemple of an Autojob with filters applied as input : 
```
name: "person_agg_view"
views:
  person_View: "BQ:ebiz-europe-west1:person.person.comet_filter(date_of_birth > {{partition_date}} AND name = {{input_name}})"
tasks:
  - domain: "person_agg_domain"
    area: "business"
    dataset: "person_agg_kpis"
    write: "OVERWRITE"
    presql:
      - |
        create or replace temporary view test_sql AS
        SELECT * FROM person_View
    sql: |
      select name, to_date({{date_event}}, "dd-MM-yyyy") AS event_date,
      current_timestamp() AS processing_time 
      from test_sql
    partition: ["event_date"]
    sink: 
      type: "BQ"
```
In this case, our Spark Job will read only partitions that are greater than {{partition_date}} and with a specific {{input_name}}.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### Deploy Notes
Execute this command on a Dataproc cluster:
```
gcloud dataproc jobs submit spark \
  --region europe-west1 \
  --cluster=cluster-a8d5  \
  --jars=gs://comet_artifacts_amines/comet-spark3_2.12-0.1.15-SNAPSHOT-assembly.jar \
  --properties=^#^spark.hadoop.fs.defaultFS=gs://datastore-amines#spark.sql.sources.partitionOverwriteMode=DYNAMIC#spark.jars.packages=com.google.cloud.spark:spark-bigquery-with-dependencies_2.12:0.17.0#spark.driver.extraJavaOptions="-DCOMET_FS=gs://datastore-amines -DCOMET_METRICS_ACTIVE=true -DCOMET_METRICS_INDEX_TYPE=BigQuery" \
  --bucket=datastore-amines \
  --class=com.ebiznext.comet.job.Main \
  -- job --name person_agg_view --options date_event="'29-09-2020'",partition_date="'1984-12-21'",input_name="'hello world'"
```

## Contributor checklist:


src/main/resources/reference.conf
```diff
@@ -341,6 +341,7 @@ spark {
   #  sql.catalogImplementation="hive"
   #  sql.catalogImplementation="in-memory"
   sql.legacy.parquet.datetimeRebaseModeInWrite = "CORRECTED"
+  viewsEnabled = "true"
```
hayssams, line 344: gr8 we allow by default users to scan BQ views

hayssams (approved):
LGTM

Issue #355, opened 2020-09-30
[FEATURE] - Table access controls on BigQuery


AmineSagaama:
## Description
**What changes were proposed in this pull request ?**
We can share a BigQuery Table or a view with specific users, without sharing the dataset that contains it.
BigQuery Table ACL lets you set table-level permissions. Table-level permissions determine the users, groups,
and service accounts that can access a table or a view.

## Solution proposition
We can rely on BigQuery Java client library to implement this feature.
First of all, we should define a collection of bindings. A **binding** binds one or more members to a single role.
Members, can be, user accounts, service accounts, Google groups, and domains (such as G Suite).
A **role** is a named list of permissions; each role can be an IAM predefined role or a user-created role.

**YAML example**:

```
tableACL:
- members:
  - user:test@gmail.com
  - group:admins@example.com
  - domain:google.com
  - serviceAccount:composer-sa@ebiz-europe-west1.iam.gserviceaccount.com
  role: roles/bigquery.dataViewer
- members:
  - group:admins@example.com
  role: roles/bigquery.metadataViewer
- etag: BwWWja0YfJA=
- version: 3
```

- version : Specifies the format of the policy (possible values are 0, 1 and 3)
- etag : etag is used for optimistic concurrency control as a way to help prevent simultaneous updates of a policy from overwriting each other.

**BigQuery Table ACL permissions**
To set or change the access to a table or view, we have to set the IAM Policy on this Table :  

```
policy:Policy = bigquery.getIamPolicy(tableId)
Policy editedPolicy = policy
            .toBuilder()
            .addIdentity(Role.of("roles/bigquery.dataViewer"), Identity.allAuthenticatedUsers())
            .build()
updatedPolicy:Policy = bigquery.setIamPolicy(tableId, editedPolicy)
```

## Additional context
Design Doc of this PR: WIP

# Life cycle

## Complexity estimation
T-Shirt Size: M.

hayssams:
Regarding this feature. I would like to emphasize that we did not intend to support Table ACLs.
We only wanted to support Row Level Security (RLS). However Google does not intend to release RLS before Q1/Q2 next year but one of our customer need this feature by the end of October.

Hopefully, what we need this year in our project is to give access to the whole table without restriction. Since Table ACLs is going GA by the end of next month, the idea is to use it in the meantime as a workaround until RLS is released.

[...]

PR #356, opened 2020-10-01
[FEATURE] - Table access controls on BigQuery

AmineSagaama:
## Summary
What changes were proposed in this pull request ?
We can share a BigQuery Table or a view with specific users, without sharing the dataset that contains it.
BigQuery Table ACL lets you set table-level permissions. Table-level permissions determine the users, groups,
and service accounts that can access a table or a view.

We will use TABLE ACCESS CONTROLS as workaround, until RLS option is released.

Note: BigQuery table-level access control is now Generally Available : 
https://cloud.google.com/bigquery/docs/release-notes

### Example: Simple policy
Consider the following example policy that binds a member to a role (it will always be "roles/bigquery.dataViewer" for comet)

```
rls:
    - name: "rlstest"
      grants:
      - "user:test@gmail.com"
      - "user:test2@gmail.com"
      - "serviceAccount:my-service-account@project-id.iam.gserviceaccount.com"
```
In the exemple above, test@gmail.com and test2@gmail.com are granted the BigQuery Data Viewer.
So, they can Read data and metadata from the table or view.

### Policy inheritance
It's the **Union** of : 

- The policy set on the resource

- The policies set on all of resource's ancestry resource levels in the hierarchy

**Related Issue: #355**

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

## Contributor checklist:


AmineSagaama:
rebuild

src/main/scala/com/ebiznext/comet/schema/model/RowLevelSecurity.scala
```diff
@@ -9,11 +9,11 @@ package com.ebiznext.comet.schema.model
...
 case class RowLevelSecurity(
   name: String,
-  predicate: String,
-  grants: List[String]
+  predicate: String = "TRUE",
+  grants: Set[String]
 ) {
 
-  def grantees(): List[(UserType, String)] = {
+  def grantees(): Set[(UserType, String)] = {
```
mhdkassir, line 16: 👍

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
```diff
@@ -75,6 +80,40 @@ trait BigQueryJobBase extends StrictLogging {
...
+    * For each call, we compare if the existing policy is equal to the defined one (in the Yaml file)
+    * If it's the case, we do nothing, otherwise we update the Table policy
+    * @param tableId
+    * @param rls
+    * @return
+    */
+  def applyTableIamPolicy(tableId: TableId, rls: RowLevelSecurity): Policy = {
+    val BIG_QUERY_VIEWER_ROLE = "roles/bigquery.dataViewer"
+    val existingPolicy: Policy = bigquery.getIamPolicy(tableId)
+    val existingPolicyBindings: util.Map[Role, util.Set[Identity]] = existingPolicy.getBindings
```
mhdkassir, line 95: For a new table for which no IamPolicy table was set , `bigquery `returns an existingPolicy with an empty PolicyBindings ?

AmineSagaama: Not necessarily, for example, a policy set at the project level applies to the project and all resources in the project.
So, If you have an Owner role on the project, you will automatically have an owner role on the new Table.

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala
```diff
@@ -199,29 +201,41 @@ class BigQuerySparkJob(
...
+//            e.printStackTrace()
+//        }
+//      }
       SparkJobResult(None)
     }
   }
 
+  private def setTablePolicy(table: Table) = {
+    cliConfig.rls match {
+      case Some(h :: _) => applyTableIamPolicy(table.getTableId, h)
```
mhdkassir, line 234: I think we should handle the possibility of having other elements in the list which are not translated by a table policy, at least by a log message.

AmineSagaama: I wanted to use a separate case class for TableACL, and don't use the case class of RLS, because it's not the same concept.
Like you said before, I can modify the case here with something
like :
`case Some(h :: Nil) => applyTableIamPolicy(table.getTableId, h)`
Because, I prefer use a suitable case class rather than, use the RLS one, and therefore, no need to pattern match 
on the list.
What do you think ?

mhdkassir: a suitable case class would be the best solution, but if we want to stick to the RLS case class (since we are using ACL only as a workaround until RLS is available)
 `case Some (h :: Nil) `
would do the job

mhdkassir (approved):
LGTM

PR #361, opened 2020-10-08
Fixing BQ merge

mmenestret:
## Summary
When merging a source, if a BQ table exists, compare the schema extracted from the BQ table to the current accepted DF instead of reaching for the accepted parquet.
The accepted parquet might not exist for TTL reason and it should not make the ingestion fail.


AmineSagaama (approved):
LGTM

PR #362, opened 2020-10-08
Better way to handle #361

mmenestret:
## Summary
Does the same thing as #361 PR but without having to load BQ table.


mhdkassir (approved):
OK

PR #364, opened 2020-10-13
Fix Big Query Sink with scripted fields issue

pchalcol:
## Summary
At ingestion time, we have no control over the attribute `required` of the scripted fields, resulting in a schema incompatibility when sinking in Big Query, which causes the failure of the sink.
After fixing the 'required' issue of the computed (scripted) fields, it appears that data ingested from a json source does not follow the declared yaml schema columns order, which may cause issues when sinking in big query. This is fixed by forcing the schema at ingestion time for json sources

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
* Create spark schema at runtime from the yaml file
* Apply this schema to the dataframes (from parquet, bq or json)

### How has this been tested?
* unit tests
* integration tests
* ingestions on GCP with big query sinks

### Other changes
Apply the declared yaml schema in json ingestion

## Contributor checklist:

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -146,18 +146,21 @@ trait IngestionJob extends SparkJob {
                                         }
                                       } else acceptedDF).drop(Settings.cometInputFileNameColumn)
 
-    logger.info("Accepted Dataframe schema right after adding computed columns")
-    acceptedDfWithScriptFields.printSchema()
-    acceptedDfWithScriptFields.show(10)
+    logger.whenDebugEnabled {
+      logger.debug("Accepted Dataframe schema right after adding computed columns")
+      acceptedDfWithScriptFields.printSchema()
+    }
 
     val withScriptFieldsDF =
       session.createDataFrame(
         acceptedDfWithScriptFields.rdd,
         schema.sparkTypeWithRenamedFields(schemaHandler)
       )
-    logger.info("Accepted Dataframe schema before optional merge")
-    withScriptFieldsDF.printSchema()
-    withScriptFieldsDF.show(10)
+
+    logger.whenDebugEnabled {
+      logger.debug("Accepted Dataframe schema before optional merge")
+      withScriptFieldsDF.printSchema()
+    }
 
     val mergedDF = schema.merge
       .map { mergeOptions =>
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -119,35 +119,42 @@ trait IngestionJob extends SparkJob {
     val writeMode = getWriteMode()
     val acceptedPath = new Path(DatasetArea.accepted(domain.name), schema.name)
 
-    val acceptedDfWithscriptFields = (if (schema.attributes.exists(_.script.isDefined)) {
-
-                                        schema.attributes.foldRight(acceptedDF) {
+    val acceptedDfWithScriptFields = (if (schema.attributes.exists(_.script.isDefined)) {
+                                        schema.attributes.foldLeft(acceptedDF) {
                                           case (
-                                              Attribute(
-                                                name,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                _,
-                                                Some(script)
-                                              ),
-                                              df
+                                                df,
+                                                Attribute(
+                                                  name,
+                                                  _,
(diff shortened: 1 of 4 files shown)


src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -119,82 +120,95 @@ trait IngestionJob extends SparkJob {
...
-                                            )
-                                          case (_, df) => df
+                                            df.T(s"SELECT *, $script as $name FROM __THIS__")
+                                          case (df, _) => df
                                         }
-
                                       } else acceptedDF).drop(Settings.cometInputFileNameColumn)
 
+    logger.info("Accepted Dataframe schema right after adding computed columns")
+    acceptedDfWithScriptFields.printSchema()
```
AmineSagaama, line 150: maybe just for testing ?

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -119,82 +120,95 @@ trait IngestionJob extends SparkJob {
...
+    acceptedDfWithScriptFields.show(10)
+
+    val withScriptFieldsDF =
+      session.createDataFrame(
+        acceptedDfWithScriptFields.rdd,
+        schema.sparkTypeWithRenamedFields(schemaHandler)
+      )
+    logger.info("Accepted Dataframe schema before optional merge")
+    withScriptFieldsDF.printSchema()
```
AmineSagaama, line 159: maybe just for testing ?

src/main/scala/com/ebiznext/comet/schema/model/Schema.scala
```diff
@@ -100,6 +100,32 @@ case class Schema(
...
+    */
+  def sparkSchemaWithoutScriptedFields(schemaHandler: SchemaHandler): StructType =
+    sparkSchemaWithCondition(schemaHandler, _.script.isEmpty)
+
+  private def sparkSchemaWithCondition(
+    schemaHandler: SchemaHandler,
+    p: Attribute => Boolean
+  ): StructType = {
+    val fields = attributes filter p map { attr =>
+      StructField(attr.rename.getOrElse(attr.name), attr.sparkType(schemaHandler), !attr.required)
```
AmineSagaama, line 124: 👍

src/main/scala/com/ebiznext/comet/job/ingest/JsonIngestionJob.scala
```diff
@@ -106,7 +106,13 @@ class JsonIngestionJob(
     val rejectedRDD: RDD[String] =
       checkedRDD.filter(_.isLeft).map(_.left.get.mkString("\n"))
 
-    val acceptedDF = session.read.json(session.createDataset(acceptedRDD)(Encoders.STRING))
+    val appliedSchema = schema
+      .sparkSchemaWithoutScriptedFields(schemaHandler)
+      .add(StructField(Settings.cometInputFileNameColumn, StringType))
```
AmineSagaama, line 111: Do we need to specify if the column is nullable or not ?

AmineSagaama (commented):
I left you some notes, LGTM for the rest.

src/main/scala/com/ebiznext/comet/job/ingest/JsonIngestionJob.scala
pchalcol: By default the column is nullable, which is what we want. That's why I didn't specify an explicit value to the nullable attribute.

src/main/scala/com/ebiznext/comet/schema/model/Schema.scala
mhdkassir: why  `!att.required `  ? and not simply `attr.required`

mhdkassir: why `!att.required`  ? and not simply `attr.required`

mhdkassir (changes requested):

src/main/scala/com/ebiznext/comet/schema/model/Schema.scala
AmineSagaama: I asked the same question, but it turns out that is the current behavior in comet.

mhdkassir: by default `required `is set to `true `in `Attribute`, but when we are using `attr.required` to calculate the spark schema shouldn't we take the value of `attr.requried` as is?

pchalcol: if `required` is true in `Attribute`, that means that `nullable` is `false` in StructField
so `StructField.nullable` = !`Attribute.required`

mhdkassir (approved):
LGTM , just consider the comments of Amine regarding the testing logs and show(10)

PR #369, opened 2020-10-20
Bugfix: script fields with Position Format

mhdkassir:
## Summary
Correction to a problem encountered upon the ingestion of a Position format schema that has script fields.

**PR Type: Bug Fix**

**Status: Ready To Review**

**Breaking change? No**

### Solution
In `PositionIngestionJob`, loading input data is now done with a schema that does not contain the script fields. This behavior was already implemented in `DsvIngestionJob`

### How has this been tested?
Updating existing unit tests to cover the case of ingesting Position schema with a script field. 

### Other changes
- refactoring of the extraction of attributes without script fields into a `lazy val` in `Schema`
- minor correction in the generation of pre-encryption yml

### Deploy Notes

## Contributor checklist:


hayssams (approved):
LGTM

PR #374, opened 2020-10-23
simply privacy in Attribute

mhdkassir:
## Summary
Simplify the `PrivacyLevel` modeling in `Attribute`. No need to have an `Option[]` for `Attribute.privacy `since the absence of privacy level can be expressed by `PrivacyLevel.None`

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### Solution
Replace `privacy: Option[PrivacyLevel] = None`
by `privacy: PrivacyLevel = PrivacyLevel.None`

### How has this been tested?
Existing unit tests

### Other changes
Add a `@JsonIgnore` annotation for a val in Schema to avoid having it in the output of serialization. 

## Contributor checklist:

### DIFF
--- a/src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
+++ b/src/main/scala/com/ebiznext/comet/schema/generator/Xls2Yml.scala
@@ -24,12 +24,17 @@ object Xls2Yml extends LazyLogging {
           .map { attr =>
             if (
               privacy == Nil || privacy.contains(
-                attr.privacy.getOrElse(PrivacyLevel.None).toString
+                attr.privacy.toString
               )
             )
               attr.copy(`type` = "string", required = false, rename = None)
             else
-              attr.copy(`type` = "string", required = false, privacy = None, rename = None)
+              attr.copy(
+                `type` = "string",
+                required = false,
+                privacy = PrivacyLevel.None,
+                rename = None
+              )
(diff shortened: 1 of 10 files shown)


AmineSagaama (commented):
Hello @mhdkassir 
Thanks for working on this PR.
What the expected case is here and how it is going to help ?
What's the advantage of replacing `Option[PrivacyLevel] = None
`  with `PrivacyLevel.None` ?

hayssams (approved):
LGTM

mhdkassir:
> 
> 
> Hello @mhdkassir
> Thanks for working on this PR.
> What the expected case is here and how it is going to help ?
> What's the advantage of replacing `Option[PrivacyLevel] = None ` with `PrivacyLevel.None` ?

@AmineSagaama I have come into this issue while working on another PR . The use case was to identify the Schemas which are concerned by applying a privacy step (encryption) from those who are not. 
Having privacy defined as an Option[PrivacyLevel] complicates the implementation, since we have to check if attribute.privacy is not defined but also if if attribute.privacy is defined and equals PrivacyLevel.None. Both cases implies the same information, there is no added value of having the two possibilities. It can even be a source of error. In general it is useful to have Option[T] to be able to indicate the absence of T, but in this case we don't need it since we already have PrivacyLevel.None

PR #378, opened 2020-11-09
[FEATURE] Handle Yaml Validation Properly in Jobs and Domains folder

AmineSagaama:
## Summary
We shouldn't block the user, if some yaml files are incorrect in the jobs folder or domains folder.
We should, only accept valid yaml files and log invalid yaml files with a warning message.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

## Contributor checklist:
- [ x ] My code follows the code style of this project.


hayssams (approved):
Even if I don't like the ".get" call

PR #405, opened 2020-12-09
Add comet extension

hayssams:
## Summary
The JSON Schema for Comet has been registered in Schema Store (PR for domain is validated and PR for Types and Business has just been pushed and is waiting validation).
To use YAML validation on domain, types and business files, one should rename the extension to .comet.yml
To support YAML validation, Business files required a new top level tag "transform".
This top level tag is optional and useful only if one want to use YAML validation.
Writing Business Files without the top level "transform" tag is now  deprecated and may be later removed once all projects have moved to the new syntax.

Only test files have been updated in the source code to reflect the new files extension.
The SchemaHandler.loadJbFromFile has a been added to support both the presence / absence of thenew AutoJob top level tag

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


src/main/scala/com/ebiznext/comet/schema/handlers/SchemaHandler.scala
```diff
@@ -46,21 +46,24 @@ class SchemaHandler(storage: StorageHandler)(implicit settings: Settings) extend
...
+    * Types redefined in the file "types.comet.yml" supersede the ones in "default.comet.yml"
     */
   lazy val types: List[Type] = {
     def loadTypes(filename: String): List[Type] = {
-      val typesPath = new Path(DatasetArea.types, filename)
-      if (storage.exists(typesPath))
-        mapper.readValue(storage.read(typesPath), classOf[Types]).types
+      val deprecatedTypesPath = new Path(DatasetArea.types, filename + ".yml")
+      val typesCometPath = new Path(DatasetArea.types, filename + ".comet.yml")
+      if (storage.exists(typesCometPath))
```
AmineSagaama, line 57: What we do, if we have both extensions in the same directory ?

AmineSagaama (approved):

src/main/scala/com/ebiznext/comet/schema/handlers/SchemaHandler.scala
hayssams: It tries to find the new extension first. If not found it tries the new one. This allows us to stay backward compatible.

PR #441, opened 2021-01-28
Handle null values as null not empty strings

hayssams:
Handle null values as null not empty strings
**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? Yes**
PrivacyEngine.crypt takes a Map[String, Option[String]] instead of Map[String, String].


mhdkassir (commented):
LGTM

mhdkassir (approved):
LGTM

PR #443, opened 2021-01-29
Log success and errors during sink at load and transform steps.

hayssams:
## Summary
An audit log record is generated when sinking datasets. 
**Related Issue: #IssueId**

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? Yes**
If using JDBC Sinks, the schema must be updated and a "step" column added too the audit table.

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -283,7 +283,6 @@ trait IngestionJob extends SparkJob {
     (savedDataset, acceptedPath)
   }
 
-
   private[this] def sink(mergedDF: DataFrame): Try[Unit] = {
     Try {
       val sinkType = metadata.getSink().map(_.`type`)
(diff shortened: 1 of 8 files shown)


src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -210,91 +244,128 @@ trait IngestionJob extends SparkJob {
...
-          createDisposition = createDisposition,
-          writeDisposition = writeDisposition,
-          location = sink.flatMap(_.location),
-          outputPartition = sink.flatMap(_.timestamp),
-          outputClustering = sink.flatMap(_.clustering).getOrElse(Nil),
-          days = sink.flatMap(_.days),
-          requirePartitionFilter = sink.flatMap(_.requirePartitionFilter).getOrElse(false),
-          rls = schema.rls
+        SparkAuditLogWriter.append(session, log)
+      case Failure(exception) =>
```
mhdkassir, line 265: we are trying to catch the failure when sinking which depends on the result of the sink() method, but in some cases in the sink method the failure of the sink step does not throws an exception , example for BQ : 
```
          val res = new BigQuerySparkJob(config, Some(schema.bqSchema(schemaHandler))).run()
          res match {
            case Success(_) => ;
            case Failure(e) => logger.error("BQLoad Failed", e)
          }
```
The case Failure is handled by logging an error but the exception is not thrown and hence the result of sink() is not affected. This applies for the case of a JDBC Sink

hayssams: Done

mhdkassir (approved):
LGTM

PR #445, opened 2021-02-02
Scala 2.11 compatible

hayssams:
## Summary
Make comet scala 2.11 compatible for older versions of HortonWorks

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


.github/workflows/snapshot.yml
```diff
@@ -21,4 +21,4 @@ jobs:
       - name: Publish Snapshot Spark 3
         run: COMET_SPARK_VERSION=3.1.0 sbt ++2.12.12 publish
       - name: Publish Snapshot Spark 2
-        run: COMET_SPARK_VERSION=2.4.7 sbt ++2.12.12 publish
+        run: COMET_SPARK_VERSION=2.4.7 sbt +publish
```
cchepelov, line 24: which version of scala are we publishing here ? Shouldn't we make explicit if we're publishing 2.11.12 ? 

If it is the case, shouldn't we _also_ keep publishing 2.12.12 + Spark 2.4?

build.sbt
```diff
@@ -45,13 +42,10 @@ libraryDependencies ++= {
...
       case "2" =>
         CrossVersion.partialVersion(scalaVersion.value) match {
-          case Some((2, scalaMinor)) if scalaMinor == 12 => (spark_2d4_forScala_2d12, jackson212)
-          case Some((2, scalaMinor)) if scalaMinor == 11 =>
-            sparkMinor match {
-              case "1" => (spark_2d1_forScala_2d11, jackson211)
-              case _   => (spark_2d4_forScala_2d11, jackson211)
-            }
+          case Some((2, 12)) => (spark_2d4_forScala_2d12, jackson212)
+          case Some((2, 11)) => (spark_2d4_forScala_2d11, jackson211)
```
cchepelov, line 46: (implicitly, I'm reading here that we keep spark 2.4 + Scala 2.11 and clean up debris from Spark 2.11 + Spark 2.1 which we hadn't published in a while, is this correct?)

project/Common.scala
```diff
@@ -129,8 +128,9 @@ object Versions {
   val h2 = "1.4.200" // Test only
   val poi = "5.0.0"
   val scalate = "1.9.6"
-  val akkaHttp = "10.2.3"
-  val akkaStream = "2.6.12"
+  val akkaHttp = "10.1.9"
```
cchepelov, line 131: 10.1.13 is published for all of scala 2.11, 2.12 and 2.13; shouldn't we use that?

project/Common.scala
```diff
@@ -129,8 +128,9 @@ object Versions {
...
+//  val akkaStream = "2.6.12"
+  val akkaStream = "2.5.9"
```
cchepelov, line 133: The last Akka-stream 2.5 is 2.5.32 , shouldn't we use that for Scala 2.11 / 2.6.12 in scala > 2.11 ?

cchepelov (commented):
Slightly concerned by the message sent by dropping scala 2.12 + Spark 2.4 while reviving scala 2.11 if this is the case

this said, with the minor nitpicks, LGTM

project/Common.scala
hayssams: Upgraded too 10.1.13

project/Common.scala
hayssams: Upgraded to 2.5.32

.github/workflows/snapshot.yml
hayssams: Removing the explicit reference to scala version will cross compile / publish for scala 2.11 & 2.12 when spark version is 2.X

cchepelov: OK good

(we might have to be cautious when we start introducing spark 3.1 / scala 2.13 builds, if we can straddle that wide a chasm)

cchepelov (approved):
LGTM

PR #451, opened 2021-02-11
Bugfix/poi version + date parsing

mhdkassir:
## Summary
- Downgrade poi version to 4.1.2. Version 5.0.0 seems to cause a regression when using comet as a dependency. 
- Manage the parsing of dates with a pattern of year and month (e.g. yyyyMM) to a date with the first day of the month
- Minor correction when saving a csv output while the dataframe is empty

**PR Type: Bug Fix**

**Status:Ready to review**

**Breaking change?No**

## How has this been tested?
Local testing


hayssams (approved):
LGTM

PR #454, opened 2021-02-18
bug fix to include the Sink options in the deserialised Sink case class

mhdkassir:
## Summary
Bug fix on the deserialization of `Sink.options`

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
Move options to the case class level instead of abstract class
Add an Option[] to the Map with a getOptions method to set default value to Map.empty

### How has this been tested?
Unit Tests and local tests

## Contributor checklist:


hayssams (approved):
LGTM

PR #458, opened 2021-03-02
Do not compute colMap before it is used.

hayssams:
## Summary
The objective is to delay the computation of the column map until it is used in a privacy Algo.
Almost all privacy Algo do not use it. We change the function signature to a call by name 

**Related Issue: #IssueId**

**PR Type: Feature**

**Status:Ready to review**

**Breaking change? Yes**
Existing privacy subclass should now change the colMap argument to a call by name argument.


mhdkassir (approved):
LGTM

PR #524, opened 2021-05-18
Sinking to a file is now a first class citizen

hayssams:
## Summary
Historically, file sink was the only option and thus was treated a little bit differently than other sinks.

File sink while still the default is now controlled by the sink tag in the YAML file. The option sink-to-file is removed and used for testing purpose only.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


mhdkassir (commented):
LGTM

mhdkassir (approved):

Issue #562, opened 2021-07-25
[BUG] - default value not applied on null columns


mhdkassir:
## Description
default value of an attribute is not applied when the column has a _null_ value in the loaded dataFrame

## Expected behavior
the column with a null value should take the defined default value in the accepted dataFrame

## Current Behavior
the column value stays _null_

## Complexity estimation
S

Issue #561, opened 2021-07-25
[BUG] - loading empty file for schema with script fields


mhdkassir:
## Description
Loading an empty DSV data files (0 lines) fails when the schema contains script fields

## Expected behavior
Loading ends with no error with 0 input and 0 lines accepted.  
## Current Behavior
Loading fails with an error while trying to calculate script fields because of missing columns in the accepted dataframe

PR #563, opened 2021-07-25
Multiple Bug fixes

mhdkassir:
## Summary
Fixing some identified bugs

**Related Issue: #562,#561,#564**

**PR Type: Bug Fix**

**Status: Ready To Review**

**Breaking change? No**

### How has this been tested?
- new Unit Tests
- no regression on existing unit tests
- tests on GCP platform for BQ jobs

### Other changes
Minor bug fix for yaml generation from Excel files by removing _nbsp_ characters from all cells to avoid erros. 

## Contributor checklist:

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -579,7 +579,7 @@ trait IngestionJob extends SparkJob {
 
       val nbPartitions = metadata.getSamplingStrategy() match {
         case 0.0 => // default partitioning
-          if (csvOutput())
+          if (csvOutput() || dataset.rdd.getNumPartitions == 0) // avoid error for an empty dataset
             1
           else
             dataset.rdd.getNumPartitions
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -1102,11 +1102,13 @@ object IngestionUtil {
       }
     }
 
-    val colValue = trimmedColValue.map { trimmedColValue =>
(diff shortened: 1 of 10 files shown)


src/main/scala/com/ebiznext/comet/job/transform/AutoTaskJob.scala
```diff
@@ -137,9 +137,9 @@ case class AutoTaskJob(
           s"SELECT $allColumns FROM $queryExpr"
         }
       queryName + " AS (" + selectExpr + ")"
-    } mkString ("WITH ", ",", " ")
-
-    val sql = subSelects + task.getSql().richFormat(sqlParameters)
+    }
+    val subSelectsString = if (subSelects.nonEmpty) subSelects.mkString("WITH ", ",", " ") else ""
```
hayssams, line 141: +1

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -1102,11 +1102,13 @@ object IngestionUtil {
...
-      else
-        trimmedColValue
-    }
+    val colValue = trimmedColValue
+      .map { trimmedColValue =>
+        if (trimmedColValue.isEmpty) colAttribute.default.getOrElse("")
+        else
+          trimmedColValue
+      }
+      .orElse(colAttribute.default)
```
hayssams, line 1111: What if no default ?

hayssams (approved):

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
mhdkassir: `default` is an Option[String], if no default value is defined the orElse will return None which is  the value expected when the column is null with no default value.

PR #568, opened 2021-08-04
Add projectId to bigquery table identifier

amine-hmila:
## Summary
Add projectId to bigquery table identifier

**Status: Ready to review**

**Breaking change? No**

## Contributor checklist:
- [ ] I have updated the Release notes.

### DIFF
--- a/src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
+++ b/src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
@@ -56,7 +56,7 @@ trait BigQueryJobBase extends StrictLogging {
   val bigquery: BigQuery = BigQueryOptions.getDefaultInstance.getService
 
   val tableId: TableId = BigQueryJobBase.extractProjectDatasetAndTable(
-    cliConfig.outputDataset + "." + cliConfig.outputTable
+    projectId + ":" + cliConfig.outputDataset + "." + cliConfig.outputTable
   )
 
   val datasetId: DatasetId = {
@@ -68,13 +68,13 @@ trait BigQueryJobBase extends StrictLogging {
     }
   }
 
-  val bqTable = s"${cliConfig.outputDataset}.${cliConfig.outputTable}"
+  val bqTable = s"${projectId}:${cliConfig.outputDataset}.${cliConfig.outputTable}"
 
   def getOrCreateDataset(): Dataset = {
     val dataset = scala.Option(bigquery.getDataset(datasetId))
     dataset.getOrElse {
       val datasetInfo = DatasetInfo
-        .newBuilder(cliConfig.outputDataset)
+        .newBuilder(projectId, cliConfig.outputDataset)
         .setLocation(cliConfig.getLocation())
         .build
       bigquery.create(datasetInfo)
--- a/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
+++ b/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
@@ -921,7 +921,8 @@ trait IngestionJob extends SparkJob {
             .getFields
             .size() == withScriptFieldsDF.schema.fields.length
         ) {
-          val bqTable = s"${domain.name}.${schema.name}"
+          val projectId = session.conf.get("fs.gs.project.id")
+          val bqTable = s"${projectId}:${domain.name}.${schema.name}"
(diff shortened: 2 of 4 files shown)


src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
```diff
@@ -56,7 +56,7 @@ trait BigQueryJobBase extends StrictLogging {
   val bigquery: BigQuery = BigQueryOptions.getDefaultInstance.getService
 
   val tableId: TableId = BigQueryJobBase.extractProjectDatasetAndTable(
-    cliConfig.outputDataset + "." + cliConfig.outputTable
+    projectId + ":" + cliConfig.outputDataset + "." + cliConfig.outputTable
```
hayssams, line 59: cliConfig.outputDataset may already be in the form projected.datasetid

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
```diff
@@ -68,13 +68,13 @@ trait BigQueryJobBase extends StrictLogging {
     }
   }
 
-  val bqTable = s"${cliConfig.outputDataset}.${cliConfig.outputTable}"
+  val bqTable = s"${projectId}:${cliConfig.outputDataset}.${cliConfig.outputTable}"
```
hayssams, line 71: cliConfig.outputDataset may already be in the form projectid.datasetid

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
```diff
@@ -68,13 +68,13 @@ trait BigQueryJobBase extends StrictLogging {
...
 
   def getOrCreateDataset(): Dataset = {
     val dataset = scala.Option(bigquery.getDataset(datasetId))
     dataset.getOrElse {
       val datasetInfo = DatasetInfo
-        .newBuilder(cliConfig.outputDataset)
+        .newBuilder(projectId, cliConfig.outputDataset)
```
hayssams, line 77: cliConfig.outputDataset may already be in the form projectid.datasetid

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -921,7 +921,8 @@ trait IngestionJob extends SparkJob {
             .getFields
             .size() == withScriptFieldsDF.schema.fields.length
         ) {
-          val bqTable = s"${domain.name}.${schema.name}"
+          val projectId = session.conf.get("fs.gs.project.id")
+          val bqTable = s"${projectId}:${domain.name}.${schema.name}"
```
hayssams, line 925: Ooh this is where we are having an issue.
At ingestion time we need to provide the project id. That what you do here.
What if the spark job is executed in a different project. In that case it won't work right ?

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQueryJobBase.scala
amine-hmila: I tried adding the projectId in the yml domain file , 
```
  name: "${bq_project}:my_dataset"
  directory: "/my_dataset"
```
but i had this error 
```
ERROR com.ebiznext.comet.schema.handlers.SchemaHandler: Domain with name my-project-id:mydataset should respect the pattern [a-zA-Z][a-zA-Z0-9_]{1,100}
```

after adding the colons and dash to the validation regex it still seems not working as expected either 
` val dbNamePattern = Pattern.compile("[a-zA-Z:][a-zA-Z0-9_:/-]{1,100}")`

it expects to find the files under  `/datasets/pending/my-project-id:my_dataset` directory 
 
```
21/08/05 22:29:45 INFO com.ebiznext.comet.workflow.IngestionWorkflow: Watch Domain: my-project-id:my_dataset
21/08/05 22:29:45 INFO com.ebiznext.comet.workflow.IngestionWorkflow: List files in /datasets/pending/my-project-id:my_dataset
```

is that the right way to include the project id or do we still have some issues to fix ?

hayssams: You can't add a projected in the domain name
Actually you should not need to do that

hayssams:
solved by referencing project id in  YML file

Issue #571, opened 2021-08-11
Merge with timestemp does not remove duplicates


abdelhakimbendjabeur:
### Description
When loading a file into a BQ Table in an incremental way using merge options (`key `and `timestamp`), comet does not remove duplicates.
### Expected behavior
Comet should keep the most recent row in the group that shares the same `key` field using the `timestamp` field.
e.g.

actual table : 
| id  | label | updated_at |
| --- | ---- | --------------- |
| 1 | value1  | 2021-08-01  |  
| 2 | value2  | 2021-08-01  |
 
new additions
| id  | label | updated_at |
| --- | ---- | --------------- |
| 1 | value1_updated  | 2021-08-02  |
| 3 | value3  | 2021-08-02  |

expected result
| id  | label | updated_at |
| --- | ---- | --------------- |
| 1 | value1_updated  | 2021-08-02  |
| 2 | value2  | 2021-08-01  |
| 3 | value3  | 2021-08-02  |

### Current Behavior ( Steps to reproduce )
It's better to put it in a picture

![comet_bug_merge_png](https://user-images.githubusercontent.com/25367149/129063914-ad62304c-1ae0-4fb6-9430-8d800804c085.png)

The Fourth load is to show that the write disposition does not affect the load in case merge is present
### How to reproduce
You can replay the above scenarios using the included zip archive.
There is one yaml file and three different csv files.

[comet_bug.zip](https://github.com/ebiznext/comet-data-pipeline/files/6969905/comet_bug.zip)

### How to fix
Should be around https://github.com/ebiznext/comet-data-pipeline/blob/master/src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala#L822

PR #572, opened 2021-08-12
fix merge with timestamp (#571)

seyguai:
## Summary
The merge algorithm isn't valid when timestamp is set: it only removes from the existing dataset, not the incoming one.

**Related Issue: #571**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
New algorithm :
- identifies the keys that have duplicated rows from both the incoming & the existing dataset
- extract the rows that match those keys
- sort those rows using the timestamp column, keep all but the latest entry for each windows (the unwanted duplicates)
- remove those rows from the union of the two datasets

### How has this been tested?
Updating an existing unit test.

## Contributor checklist:


src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -809,40 +808,54 @@ trait IngestionJob extends SparkJob {
...
-
     // We remove from the incoming data the records to delete
     val updatesDF = merge.delete
       .map(condition => in.filter(s"not ($condition)"))
       .getOrElse(in)
+
+    val (toDeleteDF, mergedDF) = merge.timestamp
+      .map { timestamp =>
+        // We only keep the first occurrence of each record, from both datasets
+        val orderingWindow =
```
hayssams, line 819: We order the records by having the last record first

hayssams (approved):
LGTM

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -809,40 +808,54 @@ trait IngestionJob extends SparkJob {
...
+          .join(keysWithDuplicatesDF, merge.key)
+          .select(allRowsDF.columns.map(col): _*)
+
+        val duplicatedRowsDF = rowsWithDuplicatesDF
+          .withColumn("rownum", row_number.over(orderingWindow))
+          .where(col("rownum") =!= 1)
+          .drop("rownum")
+
+        (duplicatedRowsDF, allRowsDF.except(duplicatedRowsDF))
+      }
```
abdelhakimbendjabeur, line 839: Why not just do 
```scala
val (toDeleteDF, mergedDF) = merge.timestamp
  .map { timestamp =>
    // Put everything in one DF
    val allRowsDF = existing.union(updatesDF)
  
    // Define window
    val orderingWindow =
          Window.partitionBy(merge.key.head, merge.key.tail: _*).orderBy(col(timestamp).desc)

    // Deduplicate
    val mergedDF= allRowsDF
      .withColumn("rownum", row_number.over(orderingWindow ))
      .where(col("rownum") === 1)
      .drop("rownum")

    // Compute rows that will be deleted for logging purposes apparently
    val toDeleteDF = rowsWithDuplicatesDF
      .withColumn("rownum", row_number.over(orderingWindow))
      .where(col("rownum") =!= 1) 
      .drop("rownum")

     (toDeleteDF, mergedDF)
  }
```

hayssams: LGTM

hayssams (approved):

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
hayssams: @seyguai Feel free to merge anytime

PR #575, opened 2021-08-24
Support Query pushdown

hayssams:
## Summary
Using the task.engine YAML tag we are now able to push the entire query to BigQuery.
Previously we were able to push the entire job including the sink action to the backend.
This feature allow us to keep spark doing the sinking while delegating the query execution to the backend engine. This is useful when we want use dynamic parititioning currently not supported by some backend like BigQuery

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


mhdkassir (approved):
LGTM

PR #583, opened 2021-08-31
Bug fix/yaml to sink

seyguai:
## Summary
Fix Sink options deserialization

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
The @JsonIgnore was ignoring every option provided.
It was supposed to be a bugfix, but I found no test attached, so I added some.

### How has this been tested?
New unit test

## Contributor checklist:

### DIFF
--- a/src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
+++ b/src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
@@ -82,7 +82,11 @@ class SinkTypeDeserializer extends JsonDeserializer[SinkType] {
   *   - BQ : Dataset is sinked to BigQuery. See BigQuerySink below
   *   - None: Don't sink. This is the default.
   */
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "type")
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  include = JsonTypeInfo.As.EXISTING_PROPERTY,
+  property = "type"
+)
 @JsonSubTypes(
   Array(
     new JsonSubTypes.Type(value = classOf[FsSink], name = "FS"),
--- a/src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
+++ b/src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
@@ -20,7 +20,7 @@
 
 package com.ebiznext.comet.schema.model
 
-import com.fasterxml.jackson.annotation.{JsonIgnore, JsonSubTypes, JsonTypeInfo, JsonTypeName}
+import com.fasterxml.jackson.annotation.{JsonSubTypes, JsonTypeInfo, JsonTypeName}
 import com.fasterxml.jackson.core.JsonParser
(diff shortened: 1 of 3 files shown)


src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
```diff
@@ -96,8 +100,6 @@ sealed abstract class Sink(val `type`: SinkType) {
   def name: Option[String]
   def options: Option[Map[String, String]]
 
-  // work around for Jackson deserialization issue (https://github.com/FasterXML/jackson-module-scala/issues/87)
-  @JsonIgnore
   def getOptions: Map[String, String] = options.getOrElse(Map.empty)
```
hayssams, line 103: I think this issue is still present in Spark 2.4 with an older version of Jackson XML
We need to be a little bit conservative here and probably rewrite it as follows:
```def getOptions: Map[String, String] = if (options == nul) Map.empty else options.getOrElse(Map.empty)```
@seyguai what do you think ?

**Review by hayssams — CHANGES_REQUESTED** [review]

File: src/main/scala/com/ebiznext/comet/schema/model/Sink.scala [hunk 2]
```diff
@@ -82,7 +82,11 @@ class SinkTypeDeserializer extends JsonDeserializer[SinkType] {
   *   - BQ : Dataset is sinked to BigQuery. See BigQuerySink below
   *   - None: Don't sink. This is the default.
   */
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  include = JsonTypeInfo.As.EXISTING_PROPERTY,
```
mhdkassir, line 87: what is the advantage of using EXISTING_PROPERTY compared to PROPERTY

src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
seyguai: Is there a way to reproduce the bug, so that I can put it in the tests ?

src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
seyguai: By default, when using JsonTypeInfo, jackson appends to the serialized object an additional key/value representing the type.
In our case, the type is already serialized from the property.

> Inclusion mechanism similar to PROPERTY with respect to deserialization; but one that is produced by
> a "regular" accessible property during serialization. This means that TypeSerializer will do
> nothing, and expects a property with defined name to be output using some other mechanism (like
> default POJO property serialization, or custom serializer).

src/main/scala/com/ebiznext/comet/schema/model/Sink.scala
hayssams: Looks like it is working since options is now an option and not anymore a Collection (in the strict sense)

hayssams (approved):
LGTM

hayssams:
@seyguai thanks

Issue #584, opened 2021-08-31
[BUG] - Lock on audit.lock fails when several jobs are running on cluster


mhdkassir:
## Description
When writing audit parquet, an error is raised when trying to get the lock. This happens particularly when several jobs are running on the spark cluster

## Current Behavior
Below is the stack trace for this error
21/08/03 14:47:52 INFO com.ebiznext.comet.job.transform.AutoTaskJob: About to write resulting dataset to /datafactory/datasets/business/MONETIQUE_MR/DS_REF_CONTRAT_ACCEPTATION_ACCEPTED
21/08/03 14:48:41 INFO com.ebiznext.comet.utils.FileLock: Trying to acquire lock for file /datafactory/audit/audit.lock during -1 ms
org.apache.hadoop.fs.FileAlreadyExistsException: 'gs://gni-stb-nxs-rtl-prd-datastore/datafactory/audit/audit.lock' already exists
    at com.google.cloud.hadoop.fs.gcs.GoogleHadoopOutputStream.createChannel(GoogleHadoopOutputStream.java:81)
    at com.google.cloud.hadoop.fs.gcs.GoogleHadoopOutputStream.<init>(GoogleHadoopOutputStream.java:70)
    at com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase.create(GoogleHadoopFileSystemBase.java:598)
    at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1126)
    at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1106)
    at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:995)
    at com.ebiznext.comet.schema.handlers.HdfsStorageHandler.$anonfun$touchz$1(StorageHandler.scala:281)
    at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
    at scala.util.Try$.apply(Try.scala:213)
    at com.ebiznext.comet.schema.handlers.HdfsStorageHandler.touchz(StorageHandler.scala:281)
    at com.ebiznext.comet.utils.FileLock.getLock$1(FileLock.scala:85)
    at com.ebiznext.comet.utils.FileLock.tryLock(FileLock.scala:117)
    at com.ebiznext.comet.utils.FileLock.doExclusively(FileLock.scala:54)
[...]

PR #585, opened 2021-08-31
Handle correctly audit lock

hayssams:
## Summary
The audit lock may be deleted between the time it is detected and the time we access its last modification date.
**Related Issue: #584 **

**PR Type: Bug**

**Status: Ready to review**

**Breaking change? No**


mhdkassir (approved):
LGTM

PR #586, opened 2021-08-31
Validate on load implementation

hayssams:
## Summary
we add a new command comet validate that exit the program with an error and print error only when any other command is run. In your DAG you would need to put a extra task that run comet validate first.

We add  an extra option COMET_VALIDATE_ON_LOAD, if set would stop the process or else simply print error message.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


mhdkassir (approved):
LGTM

PR #587, opened 2021-08-31
Bug fix/load landing without extension

seyguai:
## Summary
Comet crashes when the landing zone contains a file without extension

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
Reworked the loadLanding method, using less String, more Path/Files API.

### How has this been tested?
New unit test

### Other changes
Found some errors with tests on Windows, linked to `TestHelper#applyTestFileSubstitutions`.
Fixed the pattern, but I may not have understood what was going on.

## Contributor checklist:
- [x] My code follows the code style of this project.
- [x] I have updated the Release notes.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #596, opened 2021-09-06
Fix FileSystem exception with landing raw files

seyguai:
**Related Issue: #587**

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
Back to original feature: we only use Java nio for when we deal with archives.
We also keep the new method that can recover from a FileSystem not yet installed.
This will work as long as the user provides a Java nio FileSystem implementation.

## Contributor checklist:
- [ ] I have updated the Release notes.


hayssams (approved):
LGTM

hayssams:
Thanks

Issue #582, opened 2021-08-30
[BUG] - Ingesting to BigQuery does not handle new columns


seyguai:
## Description
If we update the attributes of a schema by adding a new field, the next ingestion to BQ will fail

## Expected behavior
For instance, Comet could first update the BQ table to add the new fields (API tables.patch) before trying to process the new data.

## Current Behavior
- If a MergeOptions is set for the current schema, Comet will fail stating the columns between the existing & the new data are not matching.
- Without MergeOptions, it seems Comet simply delegates to the BQ connector the ingestion of the new data, without creating the new fields. BQ will throw an error since it can't handle the new columns.

## Steps to reproduce
1. Create a BQ table with a record with a single string field
2. Create a yml config with a Schema matching that table, and add 2 string fields: one inside the record, and one outside
3. Ingest a simple json file matching that new yaml.

[stack-comet-ingest-sink-bq-merge-with-new-fields.txt](https://github.com/ebiznext/comet-data-pipeline/files/7076870/stack-comet-ingest-sink-bq-merge-with-new-fields.txt)
[stack-comet-ingest-sink-bq-save-new-fields.txt](https://github.com/ebiznext/comet-data-pipeline/files/7076869/stack-comet-ingest-sink-bq-save-new-fields.txt)

hayssams:
@seyguai Did you try to add to the metadata section the following two attributes:

metadata:
  sink:
    options:
      allowFieldAddition: true
      allowFieldTelaxation: true

@mhdkassir is it the right syntax ?

seyguai:
I tried adding that just now:
```
    sink:
      type: "BQ"
      options:
        allowFieldAddition: "true"
        allowFieldRelaxation: "true"
```
With & without MergeOptions : same exception
By adding those options within the schema metadata or directly in the domain metadata : same exception

I'll do more tests tomorrow.

hayssams:
@mhdkassir Could you please share the Dataproc properties setting. I think that this is where you put  alllowFieldAddition & allowFieldRelaxation to true.

mhdkassir:
For ingestion with Merge , adding new fields is not supported. We have a control in `mergeFromBQ` to make sure the input and existing data have the same number of columns, else wise we throw an Exception (line 1056, IngestionJob)

[...]

PR #601, opened 2021-09-09
Merge two datasets with different schemas

seyguai:
**Related Issue: #582**

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### Solution
- check if schemas are compatible for merge (new schema contains at least all the columns from existing one, and new columns are nullable)
- produce a schema that is built from the new schema, but compatible with the existing one, in order to load the existing data without errors
- expand the existing data with new columns without any value
- do the union with the incoming data
- remove duplicated lines

### How has this been tested?
New Integration test, new unit tests

### Other changes
- rework common algorithm, split into Util class (out of IngestionJob), to simplify testing
- small refactor in BigQuerySchemaConverters (fix warnings) + bugfix for metadata description

## Contributor checklist:

### DIFF
--- a/src/main/java/com/ebiznext/comet/utils/repackaged/BigQuerySchemaConverters.java
+++ b/src/main/java/com/ebiznext/comet/utils/repackaged/BigQuerySchemaConverters.java
@@ -23,6 +23,11 @@ import com.google.cloud.bigquery.Field;
 import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.StandardTableDefinition;
+import com.google.cloud.bigquery.TableDefinition;
+import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.TimePartitioning;
+import com.google.cloud.spark.bigquery.SupportedCustomDataType;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.avro.generic.GenericRecord;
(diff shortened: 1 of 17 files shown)


src/main/scala/com/ebiznext/comet/utils/MergeUtils.scala
```diff
@@ -0,0 +1,165 @@
...
+        missingTypes.foldLeft(originalDF) {
+          (dataframe: DataFrame, missingType: (List[String], DataType)) =>
+            missingType match {
+              case (Nil, _) => dataframe
+              case (colName :: Nil, dataType) =>
+                dataframe.withColumn(colName, lit(null).cast(dataType))
+              case (colName :: tail, dataType) =>
+                dataframe.withColumn(
+                  colName,
+                  col(colName).withField(tail.mkString("."), lit(null).cast(dataType))
```
hayssams, line 154: Not sure this cover all cases on deep column match

hayssams (changes requested):
Great PR
However detailed error reporting is missing in case of incompatible schemas. It would be nice to add it by including what has been done in the following PR bug fix/merge-bq.

src/main/java/com/ebiznext/comet/utils/repackaged/BigQuerySchemaConverters.java
```diff
@@ -19,12 +19,12 @@
  * Included here only to avoid repackaged conflict
  */
 
-import com.google.common.base.Preconditions;
 import com.google.cloud.bigquery.Field;
```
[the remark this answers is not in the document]
seyguai: Why was this duplicated then ?
I updated the content based on com.google.cloud.spark.bigquery.SchemaConverters sources for v0.22.0, which fixed the bug I had wrongly fixed (not using "comment" to store metadata descriptions)
I kept fix for Timestamp legacy type.

src/main/scala/com/ebiznext/comet/utils/MergeUtils.scala
seyguai: I'm pretty sure this only works since Spark 3.1.1

src/main/java/com/ebiznext/comet/utils/repackaged/BigQuerySchemaConverters.java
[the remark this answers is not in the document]
hayssams: I can't remember but this has to be done this way.
I'll look at the commit comments and get back to you

src/main/scala/com/ebiznext/comet/utils/MergeUtils.scala
hayssams: Probably we should make it available for spark 2 by duplicating its implementation somewhere is the repackaged section
what do you think ?

hayssams:
@seyguai 
This fails on BigQuery (SchemaUpdate not allowed on WRITE_TRUNCATE). Please check how we have to inject session settings in bugfix/bq-merge 
```
 override def extraSparkConf(): Map[String, String] =
     if (cliConfig.writeDisposition == "WRITE_TRUNCATE")
       cliConfig.options ++ Map(
         "spark.datasource.bigquery.allowFieldAddition"   -> "false",
         "spark.datasource.bigquery.allowFieldRelaxation" -> "false"
       )
     else
       cliConfig.options
```

src/main/scala/com/ebiznext/comet/utils/MergeUtils.scala
seyguai: Without any proper way to test properly & maintain any duplicated code, i would rather not do that.
I used [this article](https://medium.com/@fqaiser94/manipulating-nested-data-just-got-easier-in-apache-spark-3-1-1-f88bc9003827) to provide another code that would work with Spark < 3.1.0

hayssams (approved):
LGTM

